PR #41, opened 2022-09-13
Slimmer dockerfile

recalcitrantsupplant:
Based on a Dockerfile I've used elsewhere. Uses a multi-stage build and reduces size to ~200 MB


nicholascar (approved):
Thanks @recalcitrantsupplant 

I can't judge this by eye really but it looks "good", so I'll merge in and test the script a few times.

PR #50, opened 2022-10-11
Prez refactor

recalcitrantsupplant:
This work will refactor Prez to only provide OGC compliant endpoints and *not* render HTML; which will be left to a vue JS frontend application.

Significant changes include generalising SPARQL queries to a common CONSTRUCT template; separation of SPARQL requests for context (where required, currently only for HTML) and further use of profiles to allow rendering of complex property chains, inbound links etc.


recalcitrantsupplant:
@jamiefeiss @nicholascar 

Working:
- 'data' endpoints for spaceprez and vocprez: (dataset, feature collection, feature, concept scheme, vocabulary, collection, concept, listings of all of these)
- basic unit tests for all of the data endpoints above. These compare the API response to what is in tests/data/<prez flavour>/expected_responses. @jamiefeiss if you find the API responses are missing data you need (they will be), you can update these RDF files to make the tests fail, and I can update the code accordingly @nicholascar the tests which require a SPARQL endpoint now use `conftest.py` to define environment variables; so the copying of config files is no longer necessary
- parameterised SPARQL queries for objects, listings, and annotations (labels, descriptions etc.)
- graph cache for tbox labels
- inclusion of property paths via sh:sequencePath (see example in https://github.com/RDFLib/prez/blob/3202acc6f878c30b80c8b16ae1ca918d93083a56/prez/profiles/vocprez_default_profiles.ttl#L126)
- inclusion of properties via profiles
- exclusion of properties via profiles
- inverse properties ("inbound links") via profiles
- redirection for `/object?uri=...` for spaceprez only

To do:
- CQL functionality
- test performance of concurrent annotation queries (labels etc.) vs. a single query
- non 'data' endpoints (home etc.)
- document the profile functionality
- performance test inclusion/exclusion methods (FILTER, VALUES, explicit graph pattern matches)
- extend SPARQL CONSTRUCT for listings to include a sample of members - this may require explicit specification of the 'list' property in profiles, as per example query
```
PREFIX geo: <http://www.opengis.net/ont/geosparql#>

CONSTRUCT {<https://linked.data.gov.au/dataset/facilities/education> ?p ?o}
WHERE {
  {
    <https://linked.data.gov.au/dataset/facilities/education> !rdfs:member ?o ;
                                                               ?p ?o
  }
  UNION
  { 
    SELECT * {
      <https://linked.data.gov.au/dataset/facilities/education> rdfs:member ?o ;
                                                                ?p ?o
    } LIMIT 20
  } 
} 
```
- startup caching of labels
- logging
- unify vocprez item/listing objects as per spaceprez object, if possible
- all catprez objects/functionality

jamiefeiss (changes requested):
There's still a couple of things needed for the front-end to fully function.

- All child endpoints (e.g. concept, feature collection) lack the parent's (e.g. vocab, dataset respectively) label & link
- Concept broader & narrower properties need to be included in the vocab endpoint in order to display the concept hierarchy properly
- Broader & narrower concepts need labels & links on the concept endpoint
- In the VocPrez test data, there's a concept scheme & a collection that share the same ID
- Duplicate function in vocprez router - scheme & collection
- The Alternates Profile isn't included in */profiles endpoints or in the link headers
- /c/profiles & /profiles throw server errors
- Will the home pages (/, /c, /s & /v) return any data? In the previous version these pages contained profile information. Currently the API says these endpoints don't exist or throw errors
- The only predicate with a label is dcterms:identifier, and no description & explanation props are included
    - This is probably a data issue as I am using the local SPARQL store, but I think these common ontologies should instead be included with Prez
- Concepts listed in vocprez collection endpoints lack labels & links
- dcterms:identifier datatype is still xsd:token instead of prez:slug
- For spaceprez dataset & feature collection endpoints, the members don't need to be included in the response, as the listing endpoints do that already
- (nice to have but not required) the only namespace prefixes included in turtle responses are rdfs: & xsd:, the rest are ns1:, ns2:, etc.
- Is it possible to include the label & link for each item in a header somewhere? For the front-end's alternate profiles view, having those in the headers means I could do a HEAD request rather than a GET request. This is only to get breadcrumbs to display nicely on the alternate profiles page
- It would be nice to get the API's version as well

recalcitrantsupplant:
Thanks for the list, I've checked a few off, will update here as I go

nicholascar (approved):

Issue #64, opened 2022-12-14
Enable whole-of-Prez search


nicholascar:
Part two of #63 

This issue is to be left open for whol-of-Prez search when #63 is closed.

recalcitrantsupplant:
see: https://github.com/RDFLib/prez/discussions

Issue #63, opened 2022-12-14
/search is missing


nicholascar:
I see it, https://github.com/RDFLib/prez/blob/sparql_refactor/prez/app.py#L184, but it's commented out!

We should have _at least_ a `/search` endpoint per Prez subsystem, powered up as previous prez' e.g. `/search?q=blah` for the standards VocPrez search. 

SpacePrez search could just be an endpoint alias for `/cql`

CatPrez search could be a weighted search, like VocPrez search, but for `dcat:Resource` title & description only for now (see https://github.com/surroundaustralia/vocprez-fedsearch/blob/master/fedsearch.py#L57)

Defer a whole-of-prez search (to Issue #64) for now but we should turn text indexing on for labels & descriptions for anything that has them (Concepts, Features, Resources) and then enable basic search on those fields.

edmondchuc:
Adding this for future reference - CQL spec: https://portal.ogc.org/files/96288

recalcitrantsupplant:
see: https://github.com/RDFLib/prez/discussions

PR #72, opened 2023-03-08
Update search to work based on templates

### LINKED ISSUE #70: review/document search format
and check different methods are supported

recalcitrantsupplant:
Update documentation re search
Some refactoring
resolves #70 #63 #64


jamiefeiss:
Search seems to work fine for SKOS weighted search for concepts. I noticed that SKOS weighted search is set to be the search method for all Prez flavours - is this planned to be fixed before this PR is merged?

Also, how are we expecting the search endpoints for each Prez flavour (e.g. /v/search, /s/search etc.) to behave? Right now, all search endpoints return concepts due to the SKOS weighted search, even if the non-VocPrez endpoints are being used. Should we be restricting the results to the Prez flavour those resources are from (e.g. dcat:Resources for CatPrez, etc.)?

One more thing - should we indicate what the resource type is that's returned from the search results? Right now from SKOS weighted search, there's nothing in the data that indicates that skos:Concepts are being returned.

recalcitrantsupplant:
> I noticed that SKOS weighted search is set to be the search method for all Prez flavours - is this planned to be fixed before this PR is merged

wasn't planning to but on second thought I can probably do a default 'dumb' object search that can be available for all endpoints, which just regex's over all objects in the database. This would be more appropriate than SKOS weighted search as a default. I'll add this in. (Probably a good method to disable for larger datasets though ;) )

> Should we be restricting the results to the Prez flavour those resources are from (e.g. dcat:Resources for CatPrez, etc.)?
they are restricted to a SPARQL endpoint when you hit /v/search etc. as opposed to /search which hits all available SPARQL endpoints.

The current unimplemented plan was to specify in the system profiles (i.e. prez:VocPrezProfile) what the available search methods are for that prez flavour.
I have however specified at the root API endpoint what the available search methods are (there's a reference in the dev readme)

> should we indicate what the resource type is that's returned from the search results
With the current design this would be left up to whoever implements the search method i.e. they can choose to return classes of objects along with the search results (as a CONSTRUCT query is used). I'll add it to the concept search. Although I can't think of instances where the label/object that's being searched for doesn't have a class, it's definitely possible, and I'm not sure we need to make it a hard requirement for the search queries - I would say it would be better for the UI to expect classes as optional.

recalcitrantsupplant:
@jamiefeiss see latest commit

default search is now exact match
this is specified as a search method for all prezs
the SKOS searches are only specified as a search method for VocPrez
the API will not prevent you however from running a SKOS search against SpacePrez (don't see why it should)
but it wouldn't be unreasonable for the frontend to only show the search methods that are explicitly supported for the particular prez

prez/routers/sparql.py
```diff
@@ -36,16 +39,35 @@ async def sparql(
...
+                for res in results:
+                    g = res[2]
+                    graph.__iadd__(g)
+                return await return_rdf(
+                    graph, mediatype="text/anot+turtle", profile_headers={}
+                )
     else:
         prez = pathmap.get(start_of_path)
-    mt, result = await sparql_query(query, prez)
+    _, mt, result = await sparql_query(query, prez)
```
jamiefeiss, line 67: Querying the top-level /sparql endpoint when multiple prez flavours are enabled hits an error here where the "prez" variable is unassigned.

jamiefeiss:
Thanks for that. Now all the UI needs is a prez:link for each result returned from search so users can click on the results.

recalcitrantsupplant:
I'd need to create an "item" (e.g. Spatial item, Vocab item) instance for each returned search result. This will be a bit chatty with the database (each one has to run a SPARQL query to get the identifiers), especially if there's a lot of results. Let's go with `/object?uri=` for the moment.

An alternative would be to force people to include identifiers (with datatype = prez:slug) in their search queries - including identifiers for any parent objects. I think it would be better to keep the search requirements simple though.

prez/routers/sparql.py
recalcitrantsupplant: Fixed now will add a test

prez/routers/sparql.py
```diff
@@ -36,16 +39,104 @@ async def sparql(
...
+                combined_bindings = []
+                for result in results:
+                    for binding in result[2]["results"]["bindings"]:
+                        binding["endpoint"] = {"type": "uri", "value": result[0]}
+                        combined_bindings.append(binding)
+                results[0][2]["results"]["bindings"] = combined_bindings
+                return JSONResponse(results[0][2])
+            # warning: the following is not a sensible way to return results.
+            # the user cannot know which endpoint each result came from.
+            # no current ideas for a better solution.
```
edmondchuc, line 124: We can inject a new SPARQL response var and column for each row since we know which endpoint returned the result when we query it.

Example: we inject a new var and column for each row named `system:prez:endpoint` to indicate which endpoint the results came from.

| s   | p   | o   | system:prez:endpoint                 |
|-----|-----|-----|--------------------------------------|
| `<a>` | `<b>` | `<c>` | https://example.com/vocprez/sparql   |
| `<d>` | `<e>` | `<f>` | https://example.com/catprez/sparql   |
| `<g>` | `<h>` | `<i>` | https://example.com/spaceprez/sparql |

recalcitrantsupplant: To be clear it's where an RDF mediatype is returned that I haven't added a way to know which endpoint the results came from.

Where the results are tabular application/sparql-results (picked up with the "application/json" mediatype - need to check why application/sparql-results isn't the MT) I am adding the endpoint in as an additional binding via "endpoint"

recalcitrantsupplant:
Tests passing locally. Failing here because returned mediatype = nt

PR #80, opened 2023-04-06
Fix loading remote profiles

edmondchuc:
The `profiles_graph_cache` is using the `__add__` method which returns the combined statements in a new `Graph` instead of adding the new statements into `profiles_graph_cache`. This is why the remote profiles are never loaded.

Since `profiles_graph_cache` is a global variable imported elsewhere, we are not able to update its value. This is why something like the below won't work.

```python
profiles_graph_cache += r[1]
```

Instead, we add the triples from `r[1]` directly into `profiles_graph_cache`.

### DIFF
--- a/.github/workflows/on_push_to_feature.yaml
+++ b/.github/workflows/on_push_to_feature.yaml
@@ -59,4 +59,5 @@ jobs:
           cd tests/spaceprez && poetry run pytest
           cd ../vocprez && poetry run pytest
           cd ../catprez && poetry run pytest
+          cd ../profiles && poetry run pytest
 #          cd ../local_sparql_store && poetry run pytest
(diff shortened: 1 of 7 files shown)


recalcitrantsupplant:
Should be `.__iadd__` not `__add__` - there aren't any globals used.
I've added an example test for the the profiles.
Could you please check this works for you.
I won't bother fixing the broken tests as the curie changes will mean they need updating anyway.

edmondchuc:
If we sticked to using `__add__`, we would be reassigning the new graph to the global var `profiles_graph_cache`. This is what I mean by we can't reassign/update the var's value.

Yes, we should be able to use `__iadd__` to do the same thing. Didn't realise `__iadd__` was a thing.

edmondchuc:
Can confirm using `__iadd__` adds the remote profiles to the memory graph.

PR #81, opened 2023-04-10
Curie identifiers

recalcitrantsupplant:
implements curie identifiers

all tests passing
manually checked against IDN prez

@edmondchuc could you please review + check against your GSQ data


edmondchuc:
@recalcitrantsupplant do you mind providing a summary of the new behaviour? For example, how are curies generated, the usage of vann:preferredNamespacePrefix, etc.

nicholascar (approved):

PR #84, opened 2023-04-13
Profiles fixes

recalcitrantsupplant:
- removes unused code
- uses CURIEs as profile tokens


prez/app.py
```diff
@@ -116,7 +114,7 @@ async def app_startup():
     await count_objects()
     await populate_api_info()
     await add_prefixes_to_prefix_graph()
-    await generate_support_graphs()
+    # await generate_support_graphs()
```
edmondchuc, line 117: Should probably remove this

edmondchuc (changes requested):

edmondchuc:
Just left a comment, otherwise looks good.

jamiefeiss (changes requested):
Tested with the frontend, works fine except for a couple things:

- Profiles listed in /v/profiles, /s/profiles & /c/profiles returns links for each profile as /v/profile/<id>, etc rather than /profile/<id>
- The data for each individual profile still shows dcterms:identifier as the old token rather than the curie. Is that intended behaviour?

recalcitrantsupplant:
I don't see a conceptual issue with listing the profiles at both /profiles and under any subsystems they apply to. To me, I think it's more logical to direct a user who is already at say /v/profiles to a link that adds to that path. We also list objects at both /object and under endpoints, and collections in the case of spaceprez will be at both /collections and /datasets/<ds>/collections. If this isn't causing any technical issues let's perhaps move it to a discussion and close the pr. 

Similarly for the identifiers, they don't serve any technical/system requirement now, so I think it's more a case of do they serve any other purpose. To be clear, we are not saying the dcterms:identifier is now the curie; rather the prez system identifier (only used for the URL paths and header links) is the curie. Prez now treats dcterms identifiers like any other property.

@jamiefeiss could you confirm re whether there's any system issues and we can close this/probably just discuss on signal

edmondchuc (approved):

PR #86, opened 2023-04-14
Fix rendering of VocPub SKOS Collections at /v/collection/<id>

edmondchuc:
The endpoint at `/v/collection/<path-id>` throws a KeyError when accessing the dict that contains the inbound and outbound child links. This change fixes it by updating it to the new key values.

### DIFF
--- a/prez/renderers/renderer.py
+++ b/prez/renderers/renderer.py
@@ -196,7 +196,7 @@ def generate_prez_links(graph, predicates_for_link_addition):
 
 def generate_object_endpoint_link(graph, predicates_for_link_addition):
     all_preds = (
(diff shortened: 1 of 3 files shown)


recalcitrantsupplant:
looks good - are you able to add a test so we can prevent future regressions

edmondchuc:
Trying to run the existing test suite with `pytest` and it just hangs. Is there any notes on how tests are set up and how they should be invoked?

recalcitrantsupplant:
Try
`poetry run pytest tests/spaceprez`
`poetry run pytest tests/vocprez`
etc.
I think the issue is with the 'conftest.py' - not sure

PR #96, opened 2023-04-20
David/curie bugfix

recalcitrantsupplant:
Fix bug with curie generation
Add curie generation tests


edmondchuc:
Hi @recalcitrantsupplant, just a couple of comments.

1. Do you mind providing a description on what bug you are fixing? What the behaviour was previously and what the behaviour should be?
2. Can we change the positive test to a paramtrized test?
3. Should the positive test also test the return value of the `get_curie_id_for_uri` function? Otherwise we're not testing anything besides that there are no runtime issues.

recalcitrantsupplant:
1. the bug was that a prefix failed to generate for a "short" URI; `parsed_url.path.split("/")[-2]` failed as there were not two or more parts. The URI was: https://www.slv.vic.gov.au
2. yes not sure I'll reuse those URIs elsewhere for now though
3. Yes checking there are no runtime issues is the main thing. I can check they're strings of length greater than one? any other criteria come to mind?

edmondchuc:
> yes not sure I'll reuse those URIs elsewhere for now though

Yeah, it's not really about reuse, I think it just makes it easier to understand what input caused the test to fail.

> Yes checking there are no runtime issues is the main thing. I can check they're strings of length greater than one? any other criteria come to mind?

Is the function to generate a new curie not deterministic? If it is, we can output the values in a separate script and add it as parametrized inputs here to ensure we test that the behaviour is consistent. If there are changes in behaviour, we probably need to make it a major release or at least state in the PRs and release notes of breaking changes.

recalcitrantsupplant:
> Is the function to generate a new curie not deterministic? If it is, we can output the values in a separate script and add it as parametrized inputs here to ensure we test that the behaviour is consistent. If there are changes in behaviour, we probably need to make it a major release or at least state in the PRs and release notes of breaking changes.

No it's not deterministic in the same way compute_qname isn't, if a prefix is already used, another must be generated. It's only deterministic if "vann:preferredNamespacePrefix" is specified.

edmondchuc (approved):

PR #98, opened 2023-04-21
Container images target amd64 and arm64 platforms

edmondchuc:


recalcitrantsupplant (approved):
Can't usefully test this i think? Assume it's the same as the other action

PR #102, opened 2023-04-28
Update VocPub profile to display item status, publisher and derivation mode. Fix listing pagination, add listing to include predicate statements and add support for multiple sh:sequencePaths

edmondchuc:
Prez's VocPub profile now retrieves:
- `dcterms:publisher`
- `reg:status`
- via `prov:qualifiedDerivation`, `prov:hadRole`
- via `prov:qualifiedDerivation`, `prov:entity`

Example result for Borehole Status vocabulary in Prez listing API response:

```turtle
cs: a skos:ConceptScheme ;
    dcterms:provenance "Compiled by the Geological Survey of Queensland for the Government Geoscience Information Committee (GGIC) in a Borehole information harmonisation initiative in 2023" ;
    dcterms:publisher <https://linked.data.gov.au/org/gsq> ;
    reg:status agldwgstatus:experimental ;
    skos:prefLabel "Borehole Status"@en ;
    prov:qualifiedDerivation [ prov:entity <https://linked.data.gov.au/def/site-status> ;
            prov:hadRole <https://linked.data.gov.au/def/vocab-derivation-modes/subsetting-and-extension> ] ;
    prez:link "/v/vocab/def:borehole-status" .
```

Fixes and enhancements to the listing construct query:
- Support multiple `sh:sequencePath` for each `sh:NodeShape`
- Support "include predicates" using `sh:path` within an `sh:NodeShape`
- Fix pagination in listing construct query by moving the limit and offset into a subquery

### DIFF
--- a/prez/reference_data/profiles/vocprez_default_profiles.ttl
+++ b/prez/reference_data/profiles/vocprez_default_profiles.ttl
@@ -11,6 +11,7 @@ PREFIX sh: <http://www.w3.org/ns/shacl#>
 PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
 PREFIX reg: <http://purl.org/linked-data/registry#>
 PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+PREFIX prov: <http://www.w3.org/ns/prov#>
 
 
 prez:VocPrezProfile
@@ -103,7 +104,21 @@ prez:VocPrezProfile
         a sh:NodeShape ;
         sh:targetClass skos:Concept ;
         altr-ext:outboundParents skos:inScheme ;
-    ]
+    ] ;
+    altr-ext:hasNodeShape [
+        a sh:NodeShape ;
+        sh:targetClass prez:SchemesList ;
+        sh:path dcterms:publisher, reg:status ;
+        sh:sequencePath (
+            prov:qualifiedDerivation
+            prov:hadRole
+        ) ;
+    ] ;
+    altr-ext:hasNodeShape [
+        a sh:NodeShape ;
+        sh:targetClass prez:VocPrezCollectionList ;
+        sh:path skos:definition ;
+    ] ;
 .
 
 <https://schema.org>
--- a/prez/reference_data/profiles/vocprez_default_profiles.ttl
+++ b/prez/reference_data/profiles/vocprez_default_profiles.ttl
@@ -113,6 +113,10 @@ prez:VocPrezProfile
             prov:qualifiedDerivation
             prov:hadRole
         ) ;
+        sh:sequencePath (
+            prov:qualifiedDerivation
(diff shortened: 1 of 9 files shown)


recalcitrantsupplant:
The spaceprez failure is on geo:FeatureCollections. At present for listings of objects, a single CONSTRUCT query is generated (lets call it listing query). For individual objects, two CONSTRUCT queries are generated. One for the properties (paths to include/exclude, sequencepath, inversepath), lets call it object query, and a second query for any properties that a are membership type relationships. This second query is the same query used for listing objects i.e. the listing query. Using the listing query for objects caters well to cases like geo:FeatureCollection, where there might be thousands of members, but for the purpose of describing an instance of a geo:FeatureCollection, we would want to only include a sample of the members. The second query has LIMIT / OFFSET and we would typically want to generate internal links for members, whereas for the object query, we typically don't want either of these things. So splitting the queries has made sense to simplify/standardise query construction.

By adding the "path" inclusion to the listing query, there's now duplicate querying of the paths when a user requests information for an individual object. This isn't noticeable (but still bad), except for blank nodes.

I think we can pass context to listing query about whether it's being called for an individual object, in which case, ignore direct properties, the object query will handle these; or for a listing of objects, in which case the property predicates can be used. Happy to have a go at this

edmondchuc:
Yes, please have a go at this. I tried to debug it and figure out what was going on and it was a bit hard follow, even when stepping through with a debugger.

Can you explain why it's creating the listing query when it's a request to an object?

recalcitrantsupplant:
> Yes, please have a go at this. I tried to debug it and figure out what was going on and it was a bit hard follow, even when stepping through with a debugger.
> 
> Can you explain why it's creating the listing query when it's a request to an object?

The user story would be "describing an object may include listing objects related to the focus object" e.g. when describing a Vocabulary, including the concepts under it, or when describing a Feature Collection, the Features that are members of it. Related design considerations were:
- There may be a large number of related objects / users may/may not want to include all of them when describing the focus object. In the case of a Vocabulary it is desirable to include all concepts for completeness. For a Feature Collection a sample is sufficient (there is the separate /items endpoint).
- It would be nice to have links to the related objects, as they are typically also delivered by the API
These two considerations are the same ones addressed for the listing of objects at the other API endpoints (/datasets, /collections etc.), so that query was simply reused.

recalcitrantsupplant:
do you want to run my changes with your GSQ data and advise if any issues, otherwise no issues from me

recalcitrantsupplant (approved):
approved pending Edmond check my additional changes w/ GSQ data

edmondchuc:
@recalcitrantsupplant thanks for the explanation and fixing the tests! I can confirm Prez API is responding with the new data specified in the updated VocPub profile.

PR #103, opened 2023-05-01
David/search profiles bugfixes

recalcitrantsupplant:
Search queries are 'inserted' as subqueries within the existing "object" queries; to give context to search queries. Existing profile mechanism can be used determine shape of returned search results, such as including geospatial data.
Defaults search to Jena fulltext for use with current projects
Framework for handling error codes 404, 400, 500 etc.


prez/reference_data/search_methods/search_readme.md
```diff
@@ -0,0 +1,53 @@
...
+
+- return search results bound to the `?search_result_uri` variable. This is because the search method is used as a
+sub-select in an object listing query which expects this variable.
+- accept a search term using `$TERM` in the query. This will be substituted for the search term provided by the user.
+
+Search SPARQL queries SHOULD:
+
+- include a LIMIT clause by including `$LIMIT` to limit the number of results returned. Prez will default this limit to
+20 results if a LIMIT is not specified by users.
+Example query snipped:
```
edmondchuc, line 35: No example here?

edmondchuc:
> Defaults search to Jena fulltext for use with current projects

What happens if a Prez instance is turned on without Jena FTS enabled? Does it fallback to any other search options such as the not so ideal SPARQL regex method?

edmondchuc (approved):

prez/reference_data/search_methods/search_readme.md
recalcitrantsupplant: typo will fix later today, example is further down

recalcitrantsupplant:
> > Defaults search to Jena fulltext for use with current projects
> 
> What happens if a Prez instance is turned on without Jena FTS enabled? Does it fallback to any other search options such as the not so ideal SPARQL regex method?

no not yet. it should. I think they're reasonable methods to fall back on as they're part of the SPARQL standard, so we can expect any triplestore to implement them.

the current implementation here was to unblock Jamie and provide a search capability that is functional.

people deploying prez should only enable the Jena FTS methods in profiles if Jena with FTS enabled is the backend. The current default search methods will need to be changed.

Issue #100, opened 2023-04-21
Remove deprecated usage of set-output in GitHub Actions


edmondchuc:
This is quite urgent as after May 31st 2023, GitHub will disable this functionality completely, and it will break all of our Actions that depend on it.

For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Trivial change to the new syntax though, will get around to it maybe next week.

PR #104, opened 2023-05-07
Remove deprecated GH Actions set-output usage. Update container image release tags

edmondchuc:
This fixes https://github.com/RDFLib/prez/issues/100 by removing the use of the deprecated `::set-output` GH Actions syntax.

This PR also updates how container image tags are generated.

### Container image releases

This change will continue to add the latest `major`, `major.minor` and `major.minor.patch` tags to the container image on releases.

It will no longer add the `stable` tag and instead, it will add the `latest` tag. The `latest` tag points to the latest stable release, and not the latest rolling dev builds.

For example, for a release with a git tag of `3.2.4`, the following tags will be on the container image:

- `3`
- `3.2`
- `3.2.4`
- `latest`

### Container image dev builds on push to `main` branch

Dev builds will no longer add the `latest` tag and will only contain tags constructed from the commits made on `main` branch. The dev builds will be in the form of `major.minor.{patch+1}-dev.{commits-since-last-release}.{short-commit-sha}`. This conforms to semantic versioning and will be recognised by orchestration systems to perform automatic releases.

For example, if the latest release is `3.2.4` and there has been 7 new commits since the release and the short commit SHA is `fc82562`, then the container image tag will be:

- `3.2.5-dev.7.fc82562`

### DIFF
--- a/.github/workflows/on_push_to_main.yaml
+++ b/.github/workflows/on_push_to_main.yaml
@@ -81,7 +81,9 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           push: true
-          tags: ${{ env.IMAGE_NAME }}:${{ steps.patch.outputs.patch }}-${{ steps.hash.outputs.value }}
+          tags: |
+            ${{ env.IMAGE_NAME }}:${{ steps.patch.outputs.patch }}-${{ steps.hash.outputs.value }}
+            ${{ env.IMAGE_NAME }}:dev
           # Set provenance to false due to issue documented here: https://github.com/docker/build-push-action/issues/778
           provenance: false
           platforms: linux/amd64,linux/arm64
--- a/.github/workflows/on_push_to_main.yaml
+++ b/.github/workflows/on_push_to_main.yaml
@@ -16,6 +16,54 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v3
 
+      - name: Get latest tag
+        id: latest-tag
+        uses: WyriHaximus/github-action-get-previous-tag@v1
+        with:
+          fallback: "0.1.0"
+
+      - name: Tag starts with v
+        id: tag-starts-with-v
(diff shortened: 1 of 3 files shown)


recalcitrantsupplant:
1. is there anything non standard in the GH actions used
2. looks like this doesn't change the release process? assume we still git tag and create release via GH, could you confirm?
3. could you confirm it works as intended (post merge of PR if necessary)

edmondchuc:
> is there anything non standard in the GH actions used

Not sure what is meant by non standard.

> looks like this doesn't change the release process? assume we still git tag and create release via GH, could you confirm?

Yep! Nothing has changed, it still gets triggered in the same way, which is by making a new GitHub release. It does expect a semver git tag, but I think we're doing that already anyway.

> could you confirm it works as intended (post merge of PR if necessary

Yeah I can check that things work as expected once it's merged in.

recalcitrantsupplant (approved):

jamiefeiss:
I agree with the proposed changes, except I think we should also have an "unstable" or "dev" tag for pushes to main, like Prez UI does, which doesn't have a numbered version

edmondchuc:
> I agree with the proposed changes, except I think we should also have an "unstable" or "dev" tag for pushes to main, like Prez UI does, which doesn't have a numbered version

Just to clarify, do you mean a rolling `dev` tag in addition to the dev tag proposed above? So latest pushes to `main` branch will contain a container image with two tags in the following form:
- `ghcr.io/rdflib/prez:dev`
- `ghcr.io/rdflib/prez:3.2.5-dev.7.fc82562`

edmondchuc:
@jamiefeiss - I just pushed 2 commits detailing the container image tag variants in the `README.md` and adding the `dev` tag to the latest image created on push to `main` branch.

jamiefeiss (approved):
Great, thanks

PR #105, opened 2023-05-08
minimal implmementation of update endpoint

recalcitrantsupplant:
This is a minimal implementation of prez exposing a SPARQL update endpoint.

It takes basic auth credentials supplied to one of the /c/update, /v/update, /s/update endpoints, along with a SPARQL update and passes these through to the corresponding backend SPARQL endpoints.

These changes have been minimally tested locally and against the current IDN Fuseki instance.

### DIFF
--- a/prez/sparql/methods.py
+++ b/prez/sparql/methods.py
@@ -184,21 +184,18 @@ async def sparql_construct(query: str, prez: str):
 
 
 @alru_cache(maxsize=128)
-async def sparql_update(query: str, prez: str):
-    """Returns an rdflib Graph from a CONSTRUCT query for a single SPARQL endpoint"""
-    if not query:
-        return False, None
+async def sparql_update(request, prez):
+    headers = {
+        "Authorization": request.headers.get("Authorization"),
+        "Content-Type": request.headers.get("Content-Type"),
+    }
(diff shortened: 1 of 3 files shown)


jamiefeiss:
After some testing this seems to work fine for SPARQL updates. However, there seems to be an issue with SPARQL queries, see https://github.com/RDFLib/prez-ui/issues/51

Happy to approve & merge this PR in the meantime.

prez/sparql/methods.py
```diff
@@ -184,21 +184,18 @@ async def sparql_construct(query: str, prez: str):
 
 
 @alru_cache(maxsize=128)
```
edmondchuc, line 186: Does it make sense to cache this function call when it's performing a side effect to the database over HTTP?

Imagine if the following updates are made:

- Update label with value "aaa"
- Update label with value "bbb"
- Update label with value "aaa" but this doesn't get posted to the database because it's been cached

edmondchuc (changes requested):

prez/sparql/methods.py
recalcitrantsupplant: good catch thank you

Issue #109, opened 2023-05-11
Identify in the generate_listing_construct function what are top level items


edmondchuc:
Refactor and separate the query construction for different kinds of items for easier testing and readability of the code. 

Also consider adopting Jinja2 templates instead of using nested Python f-strings for readability.

PR #110, opened 2023-05-11
Fix ordering of listing items that are "top level items"

edmondchuc:
This correctly orders the listing items by label on the server by fetching the label predicate used for the particular profile and sorts it in the inner select query. This is only applicable for things that use the `generate_listing_construct` function that *is* a "top level item" (things listed based on class).

Initially, I thought this would work across the board for all subsystem listing endpoints. However, testing the changes with the GSQ data, it looks like "top level item" is only being utilised by the VocPub profile. Therefore, the SpacePrez and CatPrez listing endpoints are not using the "top level item" logic and does not have server-side sorting applied.

If anyone knows when a thing is a "top level item" and when it isn't, please let me know. I would have thought the items in SpacePrez and CatPrez listing pages are considered top level items as well.

Here are some examples.

Listing construct query for VocPub (uses the "top level item" logic in the inner select query):

```sparql
PREFIX dcterms: <http://purl.org/dc/terms/>
PREFIX prez: <https://prez.dev/>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>

CONSTRUCT {
  ?top_level_item a <http://www.w3.org/2004/02/skos/core#ConceptScheme> .
  ?top_level_item <http://www.w3.org/ns/prov#qualifiedDerivation> ?seq_o1_0 .
  ?seq_o1_0 <http://www.w3.org/ns/prov#hadRole> ?seq_o2_0 .	
  ?top_level_item <http://www.w3.org/ns/prov#qualifiedDerivation> ?seq_o1_1 .
  ?seq_o1_1 <http://www.w3.org/ns/prov#entity> ?seq_o2_1 .
  ?top_level_item ?p ?o .        }
WHERE {
  ?top_level_item a <http://www.w3.org/2004/02/skos/core#ConceptScheme> .
  OPTIONAL { ?top_level_item ?p ?o .            
    VALUES ?p{
      <http://purl.org/dc/terms/publisher>
      <http://purl.org/linked-data/registry#status>
    } 
  }             
  OPTIONAL {
    ?top_level_item <http://www.w3.org/ns/prov#qualifiedDerivation> ?seq_o1_0 .
    ?seq_o1_0 <http://www.w3.org/ns/prov#hadRole> ?seq_o2_0 .	
    ?top_level_item <http://www.w3.org/ns/prov#qualifiedDerivation> ?seq_o1_1 .
    ?seq_o1_1 <http://www.w3.org/ns/prov#entity> ?seq_o2_1 .            
  }

  {
    SELECT ?top_level_item
    WHERE {
      ?top_level_item a <http://www.w3.org/2004/02/skos/core#ConceptScheme> .
      OPTIONAL {
        ?top_level_item <http://www.w3.org/2004/02/skos/core#prefLabel> ?label .                    
      }
    }
    ORDER BY ASC(?label)
    LIMIT 20
    OFFSET 0
  }
}
```

Listing construct query for SpacePrez. The nested select query is blank and it does not use it at all:

```sparql
PREFIX dcterms: <http://purl.org/dc/terms/>
PREFIX prez: <https://prez.dev/>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>

CONSTRUCT {
  <https://linked.data.gov.au/dataset/qldgeofeatures> ?outbound_children ?child_item .
}
WHERE {
  OPTIONAL {
  }
  <https://linked.data.gov.au/dataset/qldgeofeatures> ?outbound_children ?child_item .
  VALUES ?outbound_children { <http://www.w3.org/2000/01/rdf-schema#member> }

  {
    SELECT ?top_level_item
    WHERE {
      OPTIONAL {
      }
    }
    ORDER BY ASC(?label)
    LIMIT 20
    OFFSET 0
  }
}
```

So in summary, this PR effectively only performs server-side sorting for VocPrez. 

I think the `generate_listing_construct` function is doing too much work and it would be good if we can identify what are "top level items" and what are not, and separate these into separate functions for easier testing and understanding of the code. We can do this in a follow-up PR separate to this work. Issue created here https://github.com/RDFLib/prez/issues/109 for discussions about the follow-up PR work.

Edit: Commit https://github.com/RDFLib/prez/pull/110/commits/7f1213060a85067133b5be34c71e041bfde564b2 fixes https://github.com/RDFLib/prez/issues/111.
Edit 2: Fixes https://github.com/RDFLib/prez-ui/issues/41


nicholascar:
> it looks like "top level item" is only being utilised by the VocPub profile

So for SpacePrez, is not the "top level item" a `dcat:Dataset` and all of this could apply to that class? For CatPrez a `dcat:Resource`?

That's presumably what you thought too initially Edmond, but you say it's not the case, so I'll have to understand why its not also.

recalcitrantsupplant:
Please read the readme. I would prefer not to need to tell people this.
The first query is for a listing of items by class. i.e. a listing endpoint.
The second query is for a listing of members of an object i.e. an object endpoint.
The subquery that Edmond has added I got to work with the current test suite. As previously advised, the scenarios for when it should and should not be included should be thought through and unit tests should be added to cover all cases.
Even in cases where code has been added as an MVP I would still expect the code added has been understood.
Ordering should not be applied globally, from experience this at a minimum increases query times significantly and for large datasets with otherwise sufficient resourcing crashes triplestores. It could be added as an opt in inclusion via profiles.

recalcitrantsupplant (changes requested):
see comment

edmondchuc:
@recalcitrantsupplant you've raised a few points but it's not clear what changes you are requesting to be made to the PR.

Btw, the existing tests fail for me when I run them locally on my machine.

```
================================================================================================================================ short test summary info ================================================================================================================================
FAILED tests/sparql/test_sparql_new.py::test_generate_construct_open - TypeError: 'NoneType' object cannot be interpreted as an integer
FAILED tests/sparql/test_sparql_new.py::test_generate_listing_construct_datasets - prez.models.model_exceptions.ClassNotFoundException: No classes found for http://example.com. Prez can only display information for instances of classes
FAILED tests/sparql/test_sparql_new.py::test_generate_listing_construct_feature_collections - NameError: name 'item' is not defined
FAILED tests/sparql/test_sparql_new.py::test_generate_listing_construct_features - NameError: name 'item' is not defined
FAILED tests/sparql/test_sparql_new.py::test_generate_listing_construct_pagination - NameError: name 'item' is not defined
FAILED tests/sparql/test_sparql_new.py::test_get_profile_predicates_sequence - AssertionError: assert None == [[rdflib.term.URIRef('http://www.w3.org/2000/01/rdf-schema#member'), rdflib.term.URIRef('http://purl.org/dc/terms/identifier')]]
ERROR tests/sparql/test_sparql_new.py::test_construct_query_with_sequence
=================================================================================================================== 6 failed, 4 passed, 3 warnings, 1 error in 1.49s ====================================================================================================================
```

edmondchuc:
Tests failing locally may be related to https://github.com/RDFLib/prez/issues/121.

edmondchuc:
Just responding to the most relevant point to this PR.

> Ordering should not be applied globally, from experience this at a minimum increases query times significantly and for large datasets with otherwise sufficient resourcing crashes triplestores.

This is a known issue with SPARQL 1.1. Unless SPARQL 1.2 introduces keyset pagination using cursors or something similar, I don't see how else we can implement pagination without using `OFFSET`, `LIMIT` and `ORDER BY`.

The current pagination implementation is wrong and confusing as it returns values in some arbitrary order and is only sorted on the client-side for a subset of the overall result. I propose we get features working as they are intended, even if horribly inefficient, so that they are at least working correctly and don't cause confusion to users of Prez. If we start having datasets that are large enough to warrant investigating other methods to perform pagination, then we can tackle that issue when it comes.

hjohns:
- Sorting options to be explored: (backup with use cases)
1. sort by query - with toggle (preferred) - default sorting determined by triple used for large datasets when toggled off
2. sort in memory in Python (complex)
3. sort in the FE (undesirable)

Considerations:
- Large dataset sizes
- Smaller datasets
- Alphabetic order (which use cases need it)
- Driven by config?
- Sort by Prez flavour

Additional:
- Search function to be added on the listing pages UI

hjohns:
Target date 24th June.

edmondchuc:
Commit https://github.com/RDFLib/prez/pull/110/commits/7f1213060a85067133b5be34c71e041bfde564b2 fixes https://github.com/RDFLib/prez/issues/111.

edmondchuc:
My proposal is to get this PR merged in now. We can follow up with a separate PR to implement the config logic to disable database ordering when required by a specific project or user.

There are more pressing issues that need fixing in Prez right now.

jamiefeiss (approved):
Tested with Prez UI, looks good to me. Thanks Edmond

PR #112, opened 2023-05-15
Remove caching on SPARQL query endpoint

### LINKED ISSUE #16: Prez duplicates SPARQL bindings returned from underlying endpoint
Originally created at https://github.com/surroundaustralia/Prez/issues/56 by @cfollenf:

The SPARQL service from Prez exposing the referenced SPARQL endpoint seems to duplicate results from the latter.
Take a basic local deployment with Prez running on http://localhost:8000/ over a Fuseki running on http://localhost:3030/ds where the triple store exposes the following dataset identified as `ds`:
```
<http://my.local/catalog> rdf:type dcat:Dataset .
<http://an.other/catalog> rdf:type dcat:Dataset .
```
When lodging the SPARQL query:
`SELECT DISTINCT * WHERE { ?d a dcat:Dataset . }`
either against the http://localhost:3030/ds/query API or through Fuseki web application UI, the bindings
```
"bindings": [
      { "d": { "type": "uri" , "value": "http://an.other/catalog" } } ,
      { "d": { "type": "uri" , "value": "http://my.local/catalog" } } ]
```
are returned, as expected. When the very same query is lodged against Prez SPARQL endpoint http://localhost:8000/sparql (again, API or UI), it returns:
```
"bindings": [
            { "d": {
                    "type": "uri",
                    "value": "http://an.other/catalog" }
            }, { "d": {
                    "type": "uri",
                    "value": "http://my.local/catalog" }
            }, { "d": {
                    "type": "uri",
                    "value": "http://an.other/catalog" }
            }, { "d": {
                    "type": "uri",
                    "value": "http://my.local/catalog" } } ]
```

Possible related to https://github.com/RDFLib/prez/issues/4

Indeed disabling one of VocPrez or SpacePrez prevents SPARQL results from duplicating. Disabling both mutes the endpoint.

### LINKED ISSUE #76: Prez SPARQL result is wrong
When running a query, the result is returning with results from previous results of the same query.

I feel like it's probably a caching logic issue where we are repeatedly adding the result to the cache.

Example:

Prez instance: https://d34tz48z3oqfi8.cloudfront.net/#/sparql

Run the following query:

```sparql

SELECT (COUNT(?c) AS ?count)
WHERE {
    ?c a skos:Concept .
}
```

Each time you run the query, it will get the result back along with all the other times this query was run.

recalcitrantsupplant:
Resolves #76. Not clear on how/why the cache appends the results between requests, and apparently only on certain endpoints. Removed until the behaviour is better understood, or an alternate caching solution is introduced.

Does not resolve #16 if multiple Prez flavours are set to the same SPARQL endpoint. 

For #16 presumably Prez should deduplicate where the queries are sent - I don't want to make further changes until all of the expected behaviour is specified, or a decision is made to remove the use of multiple SPARQL endpoints.

Draft PR until above is resolved.

### DIFF
--- a/prez/sparql/methods.py
+++ b/prez/sparql/methods.py
@@ -1,5 +1,4 @@
 import logging
-import time
 from functools import lru_cache
 from typing import Dict, List, Tuple, Union
 
@@ -89,6 +88,7 @@ async def sparql_query(query: str, prez: str) -> Tuple[bool, Union[List, Dict]]:
         }
 
 
+@alru_cache(maxsize=128)
 async def sparql_construct(query: str, prez: str):
     """Returns an rdflib Graph from a CONSTRUCT query for a single SPARQL endpoint"""
     if prez == "GenericPrez":
@@ -160,6 +160,7 @@ async def sparql_update(request, prez):
         )
 
 
+@alru_cache(maxsize=128)
 async def sparql_ask(query: str, prez: str):
     """Returns an rdflib Graph from a CONSTRUCT query for a single SPARQL endpoint"""
     if not query:
@@ -196,6 +197,7 @@ async def sparql_ask(query: str, prez: str):
         }
 
 
+@lru_cache(maxsize=128)
 def sparql_ask_non_async(query: str, prez: str):
     """Returns an rdflib Graph from a CONSTRUCT query for a single SPARQL endpoint"""
     with Client() as client:
@@ -223,6 +225,7 @@ def sparql_ask_non_async(query: str, prez: str):
         }
 
 
+@lru_cache(maxsize=128)
 def sparql_construct_non_async(query: str, prez: str):
     """Returns an rdflib Graph from a CONSTRUCT query for a single SPARQL endpoint"""
 
--- a/prez/sparql/methods.py
+++ b/prez/sparql/methods.py
@@ -17,35 +17,6 @@ log = logging.getLogger(__name__)
 TIMEOUT = 30.0
 
 
-# from starlette.requests import Request
-# from starlette.responses import StreamingResponse
-# from starlette.background import BackgroundTask
-#
-#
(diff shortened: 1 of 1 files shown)


nicholascar (changes requested):
This PR seems to remove the caching from the SPARQL query endpoints but not the CONSTRUCT endpoints and seems even to _add_ cachibg to the `sparql_construct_non_async`.

Am I reading this right? Shouldn't it remove all caching?

edmondchuc:
Same question from me. Why is caching not removed for construct queries?

I think until we have a suitable cache invalidation strategy, we should remove all caching for now to ensure new data loaded into the data store is reflected in the responses of the Prez API.

recalcitrantsupplant:
There's two commits, the first removed all caching (my draft PR), the second, after our discussion added back in caching for endpoints that the sparql router doesn't use, as we didn't think there was any issue with these. I should have created a new PR to make things clearer. See https://github.com/RDFLib/prez/pull/112/files . (Looks like I did accidentally add caching to a non async endpoint too yes, this can be removed, but it's not used by the /sparql endpoint)

The short answer is the /sparql endpoint is the only endpoint that uses the sparql_query function, which now has caching removed. Every other endpoint in prez uses the other functions such as sparql_construct.

The long answer is I took the existing code and modified it minimally to get this working. I think we've conflated the http methods with the types of queries when we don't need to. I suspect we can refactor/condense the functions quite a bit and this would clarify things. E.g. I don't think we should need to choose the endpoint function based on whether it's a construct, select, ask query

edmondchuc:
Should we remove all caching until we have a way to invalidate the cache without having to restart the Prez server?

I did a small search on the codebase and it looks like the `sparql_construct` function is being used to generate the search results and listing endpoints.

PR #116, opened 2023-05-26
Generalise local dev setup

edmondchuc:
This PR improves the experience of local dev.

- Removes hardcoded path to connegp wheel
- Add a dev `main.py` entrypoint for debuggers to hook into
- Add a minimal `.env-template` which can be used as a base for `.env` and loaded in when running locally for dev work with `python-dotenv`
- The uvicorn server started in `main.py` auto reloads on code change

Assuming a supported Python version is installed (hopefully with pyenv), working on Prez should be a matter of:
- ~Create virtual environment~
- ~Install Python dependencies from requirements.* files~
- Run poetry install
- Run the dev server (`main.py`) using Python

I'll create another PR in the near future to add in `black` and `ruff` to standardise code formatting as well.

If people are into devcontainers, we can add that too to ease local dev setup and onboarding.


recalcitrantsupplant:
We're using poetry and the requirements.txt is auto generated from the pyproject.toml using the pre commit hook. There's also black formatting in the pch. The dependency resolution in poetry has been very useful and I'd prefer not to move back to requirements.txt. The auto generation of requirements.txt was done so as not to restrict prez to users who only have poetry installed. So probably easiest to remove the hard coding by incorporating the connegp code in prez or publishing it on pypi.

Not sure I personally have a need for the main.py or env files, I use IDE functionality to debug on app.py and store different configs in pycharm, but I don't think adding these will conflict either.

Could we either remove the requirement.txt changes or remove the requirement.txt files altogether. For the code formatting, I'm happy with the pch, is there a different way you'd propose doing it?

edmondchuc:
Thanks for the insights David. I think in general, we should remove the requirements.txt files. They're really designed for development, and if we are committed to using poetry, then it just makes it confusing to have both. The community of users who use Prez should really be using it through the container image published in GitHub Packages. If in the future we want to properly support Prez as a Python package, then we can package it up that way, but this doesn't require using requirements.txt files.

I've moved to using poetry for Prez and added a dev dependency and also updated to the new dev group poetry syntax. I've also removed the requirements.txt files.

I didn't realise Black was being used as I don't use pre-commit hooks and there really aren't any ways to enforce people to use them. Regardless of whether people use the pre-commit hooks or not, we really should have the checks in GitHub Actions. This way we can ensure code that's being merged adhere to our project conventions. We're missing isort or ruff to standardise imports. As said in the opening post, I'll add these in a separate PR, and enforce them in GitHub Actions.

The `main.py` file is a useful entrypoint not dependent on PyCharm, though you can still run it through PyCharm and automatically get the benefits of running uvicorn with auto-reload turned on with a configurable port. No need to have local user-specific PyCharm settings. It makes it easier to run Prez via the terminal as well with `poetry run python main.py`. The `.env` file is useful as well to be agnostic to PyCharm. It's convenient for me to configure it since I am working on different projects with Prez and often need to connect it to different Fuseki instances, etc. I sometimes also have multiple things running and need to change the port that Prez is running on. You don't have to use `main.py` or `.env` if you don't need to, it shouldn't affect your setup I think.

edmondchuc:
Small update to the README removing mentions of requirements.txt and pip usage, in favour of poetry.

edmondchuc:
Added `poetry.toml` file with `in-project = true` to create the Python virtual environment in the project directory. I prefer this setting as it's much easier to delete the virtual environment (if needed for whatever reason). If anyone has any objections on why we shouldn't do this, let me know.

recalcitrantsupplant (approved):
Looks good, don't anticipate any breaking changes so have not tested locally.

I had an issue with auto-reload previously, but it was pycharm and/or python 3.11 specific bug. If others experience this just revert the change locally.

edmondchuc:
I think I know the issue you're talking about with the auto-reload. It's probably due to you running in app.py

edmondchuc:
We should probably fix the issue of creating the `app` at the module level in the future. There may be some issues in other scenarios when we have to import the `app` object for testing and it may import multiple times and perform the side effect of re-creating the object on each import. The fix would simply be moving the `app` object initialisation into a function.

PR #107, opened 2023-05-11
Separate multiple sequence predicates into its own SPARQL OPTIONAL clause for the listing construct query

edmondchuc:
This fixes https://github.com/RDFLib/prez/issues/106.

Previous query with all the sequence predicates in a single OPTIONAL clause.

```sparql
PREFIX dcterms: <http://purl.org/dc/terms/>
PREFIX prez: <https://prez.dev/>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>

CONSTRUCT {
  ?top_level_item a <http://www.w3.org/2004/02/skos/core#ConceptScheme> .
  ?top_level_item <http://www.w3.org/ns/prov#qualifiedDerivation> ?seq_o1_0 .
  ?seq_o1_0 <http://www.w3.org/ns/prov#hadRole> ?seq_o2_0 .       
  ?top_level_item <http://www.w3.org/ns/prov#qualifiedDerivation> ?seq_o1_1 .
  ?seq_o1_1 <http://www.w3.org/ns/prov#entity> ?seq_o2_1 .
  ?top_level_item ?p ?o .        
}
WHERE {
  ?top_level_item a <http://www.w3.org/2004/02/skos/core#ConceptScheme> .
  OPTIONAL { 
    ?top_level_item ?p ?o .            
    VALUES ?p{
      <http://purl.org/dc/terms/publisher>
      <http://purl.org/linked-data/registry#status>
    } 
  }             
  OPTIONAL {
    ?top_level_item <http://www.w3.org/ns/prov#qualifiedDerivation> ?seq_o1_0 .
    ?seq_o1_0 <http://www.w3.org/ns/prov#hadRole> ?seq_o2_0 .       
    ?top_level_item <http://www.w3.org/ns/prov#qualifiedDerivation> ?seq_o1_1 .
    ?seq_o1_1 <http://www.w3.org/ns/prov#entity> ?seq_o2_1 .            
  }

  {
    SELECT ?top_level_item
    WHERE {
      ?top_level_item a <http://www.w3.org/2004/02/skos/core#ConceptScheme> .
    }
    LIMIT 20
    OFFSET 0
  }
}
```

[...]

PR #117, opened 2023-05-29
Add loading of reg status and derivation mode vocabs on startup to SPARQL store

edmondchuc:
This PR loads the Registry Status vocabulary and the Vocab Derivation Modes vocabulary on startup to the SPARQL store.

Each of the vocabs are loaded into their own system graph:
- `urn:system:reg-status`
- `urn:system:vocab-derivation-modes`

Usage just like all other loaded data assumes the SPARQL store has default union turned on.

The VocPub system profile has been updated to include `sdo:color` in API responses. This colour value is used to colour the status marker in the vocab listings.

Some other bits:
- Add `tbox_cache.nt` to `.gitignore`
- This PR includes https://github.com/RDFLib/prez/pull/107 as the fix there is needed here


nicholascar (approved):

edmondchuc:
Will merge this only once https://github.com/RDFLib/prez/pull/107 is merged as it depends on it.

PR #119, opened 2023-05-30
Catalog object resource listing table metadata

edmondchuc:
CatPrez profile changes for https://github.com/RDFLib/prez-ui/issues/57.

Fixes `sh:sequencePath` in "object listing" query and aligns it with listing query changes introduced in https://github.com/RDFLib/prez/pull/107.

### DIFF
--- a/prez/reference_data/profiles/catprez_default_profiles.ttl
+++ b/prez/reference_data/profiles/catprez_default_profiles.ttl
@@ -5,6 +5,7 @@ PREFIX geo: <http://www.opengis.net/ont/geosparql#>
 PREFIX owl: <http://www.w3.org/2002/07/owl#>
 PREFIX prez: <https://prez.dev/>
 PREFIX prof: <http://www.w3.org/ns/dx/prof/>
+PREFIX prov: <http://www.w3.org/ns/prov#>
 PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
 PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 PREFIX sh: <http://www.w3.org/ns/shacl#>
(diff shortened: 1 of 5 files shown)


edmondchuc:
Not sure why changes made in https://github.com/RDFLib/prez/pull/119/commits/e299b6a445024e7a12b8984db5bd96f4c7a12a00 makes the tests pass locally but fails in GitHub Actions. Confusing!

I'm following the graph delta in the GitHub Actions failed test to work out what to remove in the test data now.

edmondchuc:
Tests are passing locally on my machine with changes in https://github.com/RDFLib/prez/pull/119/commits/188e835108558c2a605150ad9567cb028cab2817 but no idea why it's different and failing in GitHub Actions, will investigate more.

edmondchuc:
Interesting... I opened this PR in GitHub Codespaces and it is failing just like in GitHub Actions.

I also re-cloned Prez in my Ubuntu dev box and checked out this PR and now that is also failing.

I wonder why my working copy of Prez is passing. I even deleted the pytest cache and it still works. I change some data in the test and it fails as expected.

Question is, what is the difference between my working copy's environment and the other freshly cloned environments.

edmondchuc:
Ok, I've found the issue.

My working copy of Prez, the one that had the passing tests, somehow depended on the data in the file `tbox_cache.nt`. Once I deleted this file, things started failing and the errors were in line with the other test environments.

I'll create an issue for this. I'm not sure why we need a `tbox_cache.nt` file as all Prez instances and their state should be treated as ephemeral and their state should be derived from the persistent SPARQL store.

edmondchuc:
Yay, tests passing. Please review soonish as we need to get this merged, released and deployed for showcase on Friday.

edmondchuc:
The query performance is quite slow at the moment. I've rearranged the structure of the query slightly and it's brought it down from 42 seconds to less than 1 second (on my machine) in https://github.com/RDFLib/prez/pull/119/commits/5cd39bc5fe25c70134f002243d6d488d861e962e and https://github.com/RDFLib/prez/pull/119/commits/af84be23d0111b4cecdb0320a9f7c03d4d1c32b7.

Should be good to go now.

jamiefeiss (approved):
Looks good to me

Issue #124, opened 2023-06-09
Some caching issues related to labels


edmondchuc:
If I make the following addition (adding a `!` to the end of the literal) and replace the data in the RDF store.

```diff
<https://linked.data.gov.au/def/address-type> a owl:Ontology,
        skos:ConceptScheme ;
    dcterms:created "2020-03-25"^^xsd:date ;
    dcterms:creator <https://linked.data.gov.au/org/gsq> ;
    dcterms:modified "2023-03-16"^^xsd:date ;
    dcterms:provenance "Derived from  Australian Standard AS 4590.1:2017" ;
    dcterms:publisher <https://linked.data.gov.au/org/gsq> ;
-    skos:definition "The role or use of the address in relation to a party. An address may have more than one purpose."@en ;
+    skos:definition "The role or use of the address in relation to a party. An address may have more than one purpose.!"@en ;
    skos:hasTopConcept <https://linked.data.gov.au/def/address-type/business-address>,
        <https://linked.data.gov.au/def/address-type/delivery-address>,
        <https://linked.data.gov.au/def/address-type/overseas-address>,
        <https://linked.data.gov.au/def/address-type/postal-address>,
        <https://linked.data.gov.au/def/address-type/primary-property-address>,
        <https://linked.data.gov.au/def/address-type/residential-address>,
        <https://linked.data.gov.au/def/address-type/secondary-property-address> ;
-     skos:prefLabel "Address Type"@en .
+     skos:prefLabel "Address Type!"@en .
```

Prez is combining the results from the RDF store with a local cache and the result becomes:

```turtle
<https://linked.data.gov.au/def/address-type> a owl:Ontology,
        skos:ConceptScheme ;
    dcterms:created "2020-03-25"^^xsd:date ;
    dcterms:creator <https://linked.data.gov.au/org/gsq> ;
    dcterms:modified "2023-03-16"^^xsd:date ;
    dcterms:provenance "Derived from  Australian Standard AS 4590.1:2017" ;
    dcterms:publisher <https://linked.data.gov.au/org/gsq> ;
    skos:definition "The role or use of the address in relation to a party. An address may have more than one purpose.!"@en ;
    skos:hasTopConcept <https://linked.data.gov.au/def/address-type/business-address>,
        <https://linked.data.gov.au/def/address-type/delivery-address>,
        <https://linked.data.gov.au/def/address-type/overseas-address>,
        <https://linked.data.gov.au/def/address-type/postal-address>,
[...]

Issue #115, opened 2023-05-25
Merge multiple SPARQL endpoints into one


edmondchuc:

hjohns:
To do, capture impacts:
- Environment variables for each Prez instance.
- Breaking change
- Add to release notes

Issue #121, opened 2023-05-31
The `tbox_cache.nt` is problematic in dev and testing environments


edmondchuc:
Having cached data persisted to file and loaded every time Prez is started may cause confusing situations as experienced in PR https://github.com/RDFLib/prez/pull/119. My suggestion is to have a way to not load in cached data when running Prez in "dev mode". This may be a simple flag set via environment variables to say Prez is running in dev mode and to not use caching anywhere.

How is it problematic? See https://github.com/RDFLib/prez/pull/119#issuecomment-1569382324.

I think this issue here should also address and document why this `tbox_cache.nt` is needed. We should strive to keep the Prez server as a stateless and ephemeral thing. 

If we do need to materialise data, we should persist it in the SPARQL store along with a solution to invalidate and freshen the materialised data. This introduces design and infrastructure complexity to the Prez Stack, so we should only do this when we reach that point and when it is absolutely necessary.

edmondchuc:
Closed in https://github.com/RDFLib/prez/pull/127.

Issue #95, opened 2023-04-17
SPARQL query request incorrectly adds trailing slash to endpoint URL


avillar:
The `url=""` line in [methods.py](https://github.com/RDFLib/prez/blob/eb587266591a27fb79ca012a0a34c01539e26aca/prez/sparql/methods.py#L84) is adding a trailing slash to the endpoint URL, resulting in a 404 error for some queries (e.g., for a Fuseki server, `http://localhost:3030/ogc-hosted/update` works, but `http://localhost:3030/ogc-hosted/update/` does not). 

`url=be_endpoint` can be set as a workaround (although it seems a little redundant).

avillar:
This also results in 500 errors when running CONSTRUCT. Can confirm that the workaround does indeed work.

edmondchuc:
Closed in https://github.com/RDFLib/prez/pull/127.

PR #127, opened 2023-06-15
merge sparql endpoints

recalcitrantsupplant:
update sparql functions
update tests
add otherAnnotationProps to annotations
remove startup graph posting


recalcitrantsupplant:
Could I please get a fairly comprehensive review of this as the changes are significant and breaking. In particular testing with the frontend. I have tested visually with the FE.

recalcitrantsupplant:
I've just noticed /object is not working - will fix later today. This shouldn't block others reviewing the changes.

recalcitrantsupplant:
Resolves #115 .
Resolves #115 .
Resolves #124 NB addition of /tbox-cache endpoint which shows the tbox cache; also fix of /purge-tbox-cache endpoint. Persistence/reading of cache on shutdown/startup removed.

nicholascar (approved):
This was great fun to review: I was worried about the shear number of files but I got through them quickly since it's mostly code removal and simplification! Also, I'm thrilled to see your really neat functions and bonus documentation @recalcitrantsupplant! Great PR.

edmondchuc (commented):
Phew, finished! Looks good, just left some minor comments in there. I'll run it with prez-ui and test it tomorrow. Should be good to go shortly after that.

prez/sparql/methods.py
```diff
@@ -1,223 +1,132 @@
...
+)
 
-def sparql_query_non_async(query: str, prez: str) -> Tuple[bool, Union[List, Dict]]:
-    """Executes a SPARQL SELECT query for a single SPARQL endpoint"""
+client = Client(
+    base_url=settings.sparql_endpoint,
+    auth=(settings.sparql_username, settings.sparql_password)
+    if settings.sparql_username
+    else None,
+)
```
edmondchuc, line 29: Can we not use `base_url`? When relying on the client's `base_url` value as the actual URL, it adds a trailing slash. See the description of this issue in https://github.com/RDFLib/prez/pull/125.

This is the same issue as reported in https://github.com/RDFLib/prez/issues/95.

edmondchuc (changes requested):

edmondchuc:
Requested change in https://github.com/RDFLib/prez/pull/127/files/2fcd93ab6a7daf90e9bf32f4b02386c0a10750b9#r1236251610 will fix https://github.com/RDFLib/prez/pull/127.

edmondchuc:
This line needs to be removed. https://github.com/RDFLib/prez/blob/2fcd93ab6a7daf90e9bf32f4b02386c0a10750b9/prez/routers/profiles.py#L81

prez/sparql/methods.py
```diff
@@ -1,223 +1,132 @@
...
-            "code": response.status_code,
-            "message": response.text,
-            "prez": prez,
-        }
+async def sparql(request: Request):
+    """Sends a starlette Request object (containing a SPARQL query in the URL parameters) to a proxied SPARQL
+    endpoint."""
+    url = httpx.URL(query=request.url.query.encode("utf-8"))
+    rp_req = async_client.build_request(
+        request.method, url, headers=request.headers.raw, content=request.stream()
```
edmondchuc, line 83: Passing on the `Host` header here is problematic when sending requests to servers relying on virtual hosts reading the `Host` header value and matching routes on it.

Can we do a filter and either drop the `Host` header here or drop it and add the base URL of the SPARQL endpoint? Adding the base URL in the absence of a `Host` header I think is the default behaviour when it comes to machine tools such as Postman and curl.

edmondchuc (changes requested):

prez/sparql/methods.py
edmondchuc: When testing with the GSQ Prez Fuseki via the Prez UI SPARQL page, I was getting a 404 response when sending queries that were proxied through Prez due to the `Host` header.

prez/config.py
```diff
@@ -83,8 +52,7 @@ class Settings(BaseSettings):
     top_level_classes: Optional[dict]
     collection_classes: Optional[dict]
     general_classes: Optional[dict]
-    enabled_prezs: Optional[list]
-    generate_support_graphs: bool = True
+    prez_flavours: Optional[list] = ["SpacePrez", "VocPrez", "CatPrez", "ProfilesPrez"]
```
edmondchuc, line 55: What is ProfilesPrez?

prez/sparql/methods.py
recalcitrantsupplant: This is what I've just pushed:

```async def sparql(request: Request):
    """Sends a starlette Request object (containing a SPARQL query in the URL parameters) to a proxied SPARQL
    endpoint."""
    url = httpx.URL(
        url=settings.sparql_endpoint, query=request.url.query.encode("utf-8")
    )
    rp_req = async_client.build_request(
        request.method, url, headers=request.headers.raw, content=request.stream()
    )
    return await async_client.send(rp_req, stream=True)```

So would it just be dropping the Host header in addition to this?

File: prez/config.py (hunk 3 above)
Comment (recalcitrantsupplant) [reply prez/config.py:55]: Profiles/list of profiles are treated as objects the same as spatial, vocabulary, catalog data, so there's a Prez flavour for profiles which just displays profiles using the same mechanisms/code prez uses for the "standard" flavours

File: prez/sparql/methods.py (hunk 1 above)
Comment (recalcitrantsupplant) [reply prez/sparql/methods.py]: resolved in commits below

File: prez/routers/search.py [hunk 4]
```diff
@@ -26,67 +20,34 @@ async def search(
...
-        # uses the search queries as subqueries.
-        k: generate_item_construct(v, URIRef("https://w3id.org/profile/mem"))
-        for k, v in search_queries.items()
-    }
+    selected_method = determine_search_method(request)
+    if Literal(selected_method) not in search_methods.keys():
+        return PlainTextResponse(
+            f'Search method "{selected_method}" not found. Available methods are: '
+            f"{', '.join([str(m) for m in search_methods.keys()])}"
+        )
```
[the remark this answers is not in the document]
recalcitrantsupplant: resolved in commits below

prez/config.py
```diff
@@ -2,45 +2,26 @@
...
-    spaceprez_sparql_username: A username for the SpacePrez SPARQL endpoints, if required by the RDF DB
-    catprez_sparql_username: A username for the CatPrez SPARQL endpoints, if required by the RDF DB
-    vocprez_sparql_username: A username for the VocPrez SPARQL endpoints, if required by the RDF DB
-    spaceprez_sparql_password:  A password for the SpacePrez SPARQL endpoints, if required by the RDF DB
-    catprez_sparql_password: A password for the CatPrez SPARQL endpoints, if required by the RDF DB
-    vocprez_sparql_password: A password for the VocPrez SPARQL endpoints, if required by the RDF DB
+    sparql_endpoint: Read-only SPARQL endpoint for SpacePrez
+    sparql_update: Read/write SPARQL endpoint for SpacePrez
+    sparql_username: A username for the SpacePrez SPARQL endpoints, if required by the RDF DB
+    sparql_password:  A password for the SpacePrez SPARQL endpoints, if required by the RDF DB
```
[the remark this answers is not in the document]
recalcitrantsupplant: resolved in commits below

prez/config.py
edmondchuc: Ok, I'm guessing it's reusing the logic of generic Prez to list https://gsq-prez.dev.kurrawong.ai/profiles?

prez/config.py
```diff
@@ -2,45 +2,26 @@
...
-    vocprez_sparql_endpoint: Read-only SPARQL endpoint for VocPrez
-    vocprez_sparql_update: Read/write SPARQL endpoint for VocPrez
-    spaceprez_sparql_username: A username for the SpacePrez SPARQL endpoints, if required by the RDF DB
-    catprez_sparql_username: A username for the CatPrez SPARQL endpoints, if required by the RDF DB
-    vocprez_sparql_username: A username for the VocPrez SPARQL endpoints, if required by the RDF DB
-    spaceprez_sparql_password:  A password for the SpacePrez SPARQL endpoints, if required by the RDF DB
-    catprez_sparql_password: A password for the CatPrez SPARQL endpoints, if required by the RDF DB
-    vocprez_sparql_password: A password for the VocPrez SPARQL endpoints, if required by the RDF DB
+    sparql_endpoint: Read-only SPARQL endpoint for SpacePrez
+    sparql_update: Read/write SPARQL endpoint for SpacePrez
```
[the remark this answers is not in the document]
recalcitrantsupplant: resolved in commits below

demo/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
       start_period: 20s
 
   prez:
-    image: "ghcr.io/rdflib/prez:3.0.6"
+    image: "ghcr.io/rdflib/prez:3.0.10-dev.3.440d6ee"
```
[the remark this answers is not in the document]
recalcitrantsupplant: resolved in commits below

prez/sparql/methods.py
edmondchuc: I think something like this will work.

```diff
async def sparql(request: Request):
    """Sends a starlette Request object (containing a SPARQL query in the URL parameters) to a proxied SPARQL
    endpoint."""
    url = httpx.URL(url=settings.sparql_endpoint, query=request.url.query.encode("utf-8"))
    
+    headers = []
+    for header in request.headers.raw:
+        if header[0] != b'host':
+            headers.append(header)
+    headers.append((b'host', str(url.host).encode("utf-8")))
    
    rp_req = async_client.build_request(
-        request.method, url, headers=request.headers.raw, content=request.stream()
+        request.method, url, headers=headers, content=request.stream()
    )
    return await async_client.send(rp_req, stream=True)
```

prez/config.py
recalcitrantsupplant: yes correct

prez/sparql/methods.py
recalcitrantsupplant: thanks for the code - added

edmondchuc:
I think the last thing is to remove `prez_type=GenericPrez` on line 70 and `GenericPrez` on line 80 in https://github.com/RDFLib/prez/blob/f90b743ecc097e558552a3c637528ef0174f1d6c/prez/routers/profiles.py#L70-L80.

edmondchuc:
Once you've made that change in the last comment, I'll do one last test and distill the changes into dot points here so we can paste it into the release notes.

recalcitrantsupplant:
Thanks - removed now

edmondchuc:
## Fixes

- https://github.com/RDFLib/prez/issues/115
- https://github.com/RDFLib/prez/issues/124
- https://github.com/RDFLib/prez/issues/95
- https://github.com/RDFLib/prez/issues/121
- https://github.com/RDFLib/prez/issues/55

## Changes

- Prez is now read-only for a given SPARQL endpoint
  - Prez no longer loads vocabularies to the SPARQL store on startup and instead loads it into an in-memory graph object
- Simplified docker-compose local setup
- Prez no longer persists the in-memory graph cache to disk. Each Prez restart now starts in a deterministic state with a clean cache
- No longer uses base URL in HTTP client and as a result, no longer adds a trailing `/` to the URL of the requests made using the HTTP client
- For search, invalid `search_methods` values now returns a HTTP status code of `400`
- SPARQL proxy endpoint filters out the `Host` header and sets it to the base URL of the target SPARQL endpoint

## Added

- HTTP endpoint to clear in-memory graph cache
- HTTP endpoint to view in-memory graph cache
- Documentation for Prez annotation properties
- Documentation for common Prez environment variables

## Breaking

- SPARQL endpoints for each subsystems have been consolidated into one
- SPARQL Update and SPARQL HTTP Store support dropped
- Environment variables for SPARQL related config removed and simplified
  - `SPARQL_ENDPOINT`
  - `SPARQL_USERNAME`
  - `SPARQL_PASSWORD`

@recalcitrantsupplant do you agree with this list?

edmondchuc:
One last thing, can you fix the formatting for the documentation section on annotation properties please? See https://github.com/RDFLib/prez/blob/798be47cdf185cc10d2078db1de40f7fb2e1831f/README-Dev.md#annotation-properties.

recalcitrantsupplant:
Cheers, I've edited your comment to include info on "annotated SPARQL queries" under Added, and fixed the markdown table.

edmondchuc (approved):
Looking good now. Thanks @recalcitrantsupplant for addressing my comments.

PR #128, opened 2023-06-25
Object endpoint returns system links.

### LINKED ISSUE #101: Rename general_class to item_class
Based on chat with David.

recalcitrantsupplant:
Add endpoint reference data to support system link generation. Rename inbound/outbound children/parents to focus to/from child/parent terminology to align with SHACL. Rename general_class to base_class. Resolves #101


recalcitrantsupplant:
Minimal implementation of /object endpoint providing system links is  complete.

With this change, I suspect the complexity around needing to specify subject to/from child/parent *may* be also able to be removed, though it might require an extra call to the triplestore. I'd like to have a go at this first.

The changes for /object also mean link generation now happens in two ways (one for /object which uses the reference data for which endpoints can deliver which classes) and another for individual object/listing pages, which use link constructors in the pydantic models. Ideally all link generation can be moved to the new method.

PR #134, opened 2023-07-17
Add new vocabs API

edmondchuc:
Closes https://github.com/RDFLib/prez/issues/131.

Summary:
- New API for client fetching of concept schemes and concepts
- Prefix generation is set by default on startup
- Add identifier endpoints to provide resolution for IRIs to CURIEs and vice versa
- Add parametrised tests

### New API for client fetching of concept schemes and concepts

Adds a new set of APIs as described in the linked issue above to enable the ability to fetch a vocabulary of any size in an incremental manner.

### Prefix generation is set by default on startup

This PR creates prefixes by default on start up and can be disabled by setting the `DISABLE_PREFIX_GENERATION` environment variable to `false`. This default behaviour is required as the new API assumes the endpoints are stateless with no side effects.

### Add identifier endpoints to provide resolution for IRIs to CURIEs and vice versa

New identifier endpoints to provide resolution of IRIs and CURIEs.

- `/identifier/curie/{iri:path}`
  - With an IRI as input, return its identifier in the form of a CURIE.
- `/identifier/iri/{curie}`
  - With a CURIE as input, return its fully qualified IRI.

These are required to set up parameterised tests by converting IRI inputs into their CURIE identifiers to perform requests to the new vocabulary endpoints. For example, converting parameterised IRIs to perform test requests to `/v/vocab/{concept_scheme_curie}/top-concepts`.

### Add parametrised tests

Basic structure of parameterised tests for each new vocabulary endpoint added with different scenarios tested. New scenarios can be added in the form of another parameterised test.

PR #140, opened 2023-07-31
/count and vocabs-all endpoint

edmondchuc:
This PR adds two additional endpoints:

- `/count`
- `/v/vocabs/{curie}/all`

These address the ideas discussed in the Prez meeting here https://github.com/RDFLib/prez/pull/134#issuecomment-1650813708.


edmondchuc:
For some reasons, the tests hang when running it in `tests/object`. Maybe that's why they were not included in the set of tests to run in GitHub Actions.

PR #143, opened 2023-08-04
Add reg:status to vocpub profile as an annotation prop. Add rdf:type to the vocabs API and add fix to ensure profile predicates and their values get annotation values added to response

edmondchuc:
This PR:
- Adds the `reg:status` property to the VocPub profile
- Add `rdf:type` to the vocabs API responses for clarity
- ~~Add temporary fix to ensure additional predicates defined in profiles get added with their annotation properties~~ Expand the response graph with annotations

On the last point, the current issue is, returning a response with the following data does not add the annotation for `<https://linked.data.gov.au/def/reg-statuses/valid>`.

```turtle
<http://data.bgs.ac.uk/id/BeddingSurfaceStructure/TFCR> a skos:Concept ;
    rdfs:label "Crawling / Walking Tracks and Trails"@en ;
    dcterms:identifier "bddngsrfcstrctr:TFCR"^^prez:identifier ;
    ns1:status <https://linked.data.gov.au/def/reg-statuses/valid> ;
    skos:prefLabel "Crawling / Walking Tracks and Trails"@en ;
    prez:childrenCount 0 ;
    prez:link "/v/vocab/ref:BeddingSurfaceStructure/bddngsrfcstrctr:TFCR" .
```

This is because the annotations are added to the graph object first before the additional predicates from the profiles are added.

~~This is essentially a temporary fix right now because I need this functionality to be working for a project with a deadline. Will follow up in a week or two with a refactor of the [`prez.renderers.renderer.return_annotated_rdf`](https://github.com/RDFLib/prez/blob/070d8e1401087c7a1ffcafc9a0b20b8aa9fc772e/prez/renderers/renderer.py#L75C1-L110C6) function.~~

The function has been updated in this commit https://github.com/RDFLib/prez/pull/143/commits/40225bd1cbf4210bd940de2e62ed4e3e01cc5275 to expand the graph with annotations until no new statements are added.


recalcitrantsupplant:
To clarify the changes as it is a little different from the initial description now:
"annotations are added to the graph object first before the additional predicates from the profiles are added."
The annotations for labels, descriptions, provenance, and status are added after the initial RDF response. The issue was that there was no functionality to add annotations (labels in this case) for the annotations themselves.

The change here is to now look for annotations on annotated properties themselves, recursively, until now new annotations are found. This should prevent annotations returning unlabeled URIs.

I don't have time to review this locally right now but don't anticipate any issues. The extensions to the vocprez query will need to be included in refactoring the queries to a common model.

recalcitrantsupplant (approved):
approved see comment in PR

PR #144, opened 2023-08-06
Add /identifier/redirect endpoint

edmondchuc:
This PR adds a new `/identifier/redirect` endpoint that takes a query parameter `iri` and uses it to look up in the SPARQL backend a statement containing `?iri foaf:homepage ?url`. If a value for `?url` is found, it returns a redirect response redirecting to the `?url` value.

Tests previously under `tests/curies` have been moved to `tests/identifier` and the new tests for the `/identifier/redirect` endpoint have been added to the new location as well.

Note, this PR merges into existing PR https://github.com/RDFLib/prez/pull/143.


recalcitrantsupplant:
What's the motivation / use case here? Is it intended the prez FE app will use this? does this change what is expected of users to provide in their data for prez to function?

edmondchuc:
The use case is to have a redirect response generated from the data in the database in the form of `?iri foaf:homepage ?url`.

For example, a persistent IRI such as `http://data.bgs.ac.uk/id/dataHolding/13603129` is forwarded to Prez as a query parameter at `/identifier/redirect` and Prez just performs a look up in the database and returns a redirect response or 404 if no matches found.

edmondchuc:
This redirect functionality is the same as other redirect services such as W3ID and the AGLDWG PID service. The difference here is that the redirect mapping is defined in the database using `foaf:homepage`.

recalcitrantsupplant:
I understand how it works I'm more getting at why is it needed, what is the motivation to add this functionality to prez where it didn't exist before. What sort of resources would you expect to have a foaf homepage, would you be programmatically adding this for e.g. external vocabs? Is there a current use case with a project that would describe why it is needed.

edmondchuc:
The example I provided above is a BGS persistent IRI redirecting to the metadata record in GeoNetwork. Primary motivation to have functionality like this is for cases where organisations issue their own persistent identifiers where they need a mechanism to perform the redirect from the persistent identifier to the actual web resource. Rather than implementing the redirect mapping in a config specific to NGINX or Apache HTTP, having it in Prez means the mapping is a piece of data in the SPARQL store. At least in the BGS case, this is their preferred solution.

recalcitrantsupplant:
Thanks yes that clarifies it - could you add that snippet (or similar) to one of the readmes please

recalcitrantsupplant (approved):

edmondchuc:
@recalcitrantsupplant I've added a description of redirects to the `README.md` and as a separate enhancement, I've updated the code to forward on headers from the request. This will allow clients to perform content negotiation. Tests updated as well.

Issue #147, opened 2023-08-29
The drop-down list profiles are not implemented yet (JSON/CSV)


edmondchuc:
Need to do some design and research in the best way to implement this. 

- How does the profiles define the properties to generate the result?
- What is the schema for the result? Identifier and label?
- Do we need a background job of some sort for very large objects? How do we handle large objects with 10s of thousands of values?

recalcitrantsupplant:
JSON completed
CSV to do
Tests to add

hjohns:
Future tasks:
1- remove logic from pydantic into standard routes or renderer
2- system links now have classes, where a profile/subsystem combo (filter) is not provided in the future all the rdf will be returned for an object. 
3- class/media type --> matching profiles with class/media type --> preferences --> profiles/media types in preference order
4- make sure our /object page is not broken

PR #155, opened 2023-09-22
Drop-down profile implementation

edmondchuc:
Fixes: https://github.com/RDFLib/prez/issues/147.

This PR implements the drop-down profile, which allows downstream systems to populate UI components such as a `<select>`. The values can be a concept from a list of concepts in a vocabulary or a collection or a vocabulary or collection from a list of vocabularies or collections. The drop-down profile can be used to create forms or surveys that require users to choose from a predefined set of options.

Drop down profile and media type support:
- [x] JSON (`application/json`)
- [x] CSV (`text/csv)`

Views that support this profile:
- `/v/vocab` - vocab listing
- `/v/vocab/{concept_scheme_curie}` - concept scheme drop down with concepts as the items
- `/v/collection` - vocab collection listing
- `/v/collection/{collection_curie}` - collection drop down with concepts as the items

For CSV, the listing endpoints will save the file with the file name from the profile class name. For example, `/v/vocab` will save a `SchemesList.csv`. For other object endpoints, the renderer will work out the curie of the object and use it for the CSV filename. For example, `http://linked.data.gov.au/def/address-type` will become something like `df_address-type.csv`.

Note that this implementation generates the data in memory for both media types. This may be slow for very large data but has been accepted as the current behaviour. We may delegate this kind of data processing to a background job in future iterations.

Miscellaneous:
- On startup, perform health check first
- Removed some unused imports
- `/object` endpoint
  - ~Machine requests made to `/object` will use the provided media type and profile to return an appropriate response in one of the subsystems. In some cases, the resource associated with the supplied IRI may be associated with more than one subsystem. To decide which subsystem to use, a preferred ordering of the subsystem URLs are set in the config under `subsystem_url_preferred_order`. After the system URLs are sorted to their preferred order, the first system URL will be used to process the request.~
  - **The redirect has been removed and the subsystem URL preference is no longer needed. However, I am not sure what will happen when a resource requested via the `/object` endpoint has conflicting classes. I'm assuming it will just find the first profile that constrains one of the classes on the object and go with that.**


edmondchuc:
Drop-down profile support for JSON on all views working as of commit https://github.com/RDFLib/prez/pull/155/commits/26ca3c9d37b16d0f75bd40d9c3216d7f60d8e8ea.

edmondchuc:
No idea why that one single test for spaceprez is failing. The tests pass on my local machine, even when I do a clean install.

### `test_endpoints_spaceprez.py::test_dataset_listing_anot`
```
=========================== short test summary info ============================
FAILED test_endpoints_spaceprez.py::test_dataset_listing_anot - AssertionError: None
assert False
 +  where False = <bound method Graph.isomorphic of <Graph identifier=N173ab5d971ed41449408570ec0e1e115 (<class 'rdflib.graph.Graph'>)>>(<Graph identifier=N5c4352554df34fa884c7d1ac578f2c77 (<class 'rdflib.graph.Graph'>)>)
 +    where <bound method Graph.isomorphic of <Graph identifier=N173ab5d971ed41449408570ec0e1e115 (<class 'rdflib.graph.Graph'>)>> = <Graph identifier=N173ab5d971ed41449408570ec0e1e115 (<class 'rdflib.graph.Graph'>)>.isomorphic
================== 1 failed, 5 passed, 19 warnings in 18.25s ===================
```

recalcitrantsupplant (commented):
Is the change to make the /object endpoint redirect necessary? I know you mentioned the reason but could you describe here the benefit of redirecting system users? Could this functionality be provided by downstream consumers of the information e.g. what Prez UI currently does?

The previous thinking with the object endpoint was it could provide you all information known about an object + profiles. This would be a departure from that, the differences would be:
- I assume that if prez has RDF for an object but doesn't have a profile that constrains one of the classes for that object the /object endpoint will no longer provide any response?
- Users will only know other profiled views of an object exist if the profiled view they are redirected to presents these other views via links (which they probably will - but it might not be clear to a user that there are other views of the object).

prez/renderers/json_renderer.py
```diff
@@ -0,0 +1,225 @@
...
+        iri = graph.value(predicate=RDF.type, object=target_class)
+        if iri is not None:
+            return iri
+
+    raise NotFoundError(
+        f"No resource IRI found based on the constrained classes defined in {profile}."
+    )
+
+
+def _get_child_to_focus_predicates(
```
recalcitrantsupplant, line 26: Is the plan to replace the existing functions?

edmondchuc: Using existing function now.

recalcitrantsupplant (approved):
Discussed today - some follow up items for us at a later point discussed in this issue
https://github.com/RDFLib/prez/issues/147

PR #156, opened 2023-09-25
Fix get_bnode_depth edge case

edmondchuc:
The current `get_bnode_depth` function only works for resources with a single top-level blank node.

Effectively, it is counting the total number of blank nodes in the given graph closure.

This PR fixes the issue where multiple top-level blank nodes are counted incorrectly based on the description of the previous behaviour above. It counts the maximum depth of the blank nodes for a given graph closure correctly by performing a depth-first search.

A new test case `bnode_depth-2-2.ttl` has been added to illustrate the correct behaviour has been implemented now.


hjohns (approved):
Test TTL matches the multiple agent use case, can't comment on the implementation of the py changes. This issue impacted the progressive loading of concept hierarchy, does the test case check that as well?

edmondchuc:
@hjohns, yes this is what fixes it. The progressive loading of vocabs APIs call this bnode depth function, which previously calculated it wrong if you have more than 1 top-level blank node. In your case, you had two top-level blank nodes because you had two custodians described with the `prov:qualifiedAttribution` property. This is tested in the new test data added in this PR.

recalcitrantsupplant (approved):
looks good
not anticipating cyclic graphs or large bnode depths for our use cases but by the looks they could be problematic?

edmondchuc:
Cyclic graphs are not an issue here. As described in the function, it is intended to work on a resource's concise bounded description.

PR #163, opened 2023-10-12
allow port and host to be configurable when running Prez from Docker

recalcitrantsupplant:
Parameterised Uvicorn host and port settings in the Dockerfile ENTRYPOINT, providing default values to enhance configurability in deployment scenarios


recalcitrantsupplant:
can be tested with the docker compose script in demo/

hjohns (approved):
Works!

edmondchuc (approved):

PR #164, opened 2023-10-14
restructure Catprez Endpoints and add annotation listings to root endpoint

recalcitrantsupplant:
Root endpoint ("/") now includes:
- the Prez version (when run directly with Python this will display as "0.1.0.dev0" .
- the annotation predicates Prez will look for (for labels, descriptions, and provenance) to annotate data when the annotated turtle mediatype (text/anot+turtle) is requested. These are provided as an ordered list in terms of the priority a consumer should use them in if more than one predicate is present. This list is configurable via environment variables - see prez/config.
- A list of endpoint information, example shown below:
```
endpoint:catalog-listing a ont:ListingEndpoint ;
    ont:deliversClasses prez:CatalogList ;
    ont:isTopLevelEndpoint "true"^^xsd:boolean ;
    ont:baseClass dcat:Catalog ;
    ont:endpointTemplate "/c/catalogs" ;
.
```
At this point the endpoints listed is not an exhaustive list - only the endpoints which deliver "instance data" have been included as this endpoint specification as RDF is used by the application.

The Catprez endpoints now follow the same structure as Spaceprez endpoints (with one less level of hierarchy):
/c/catalogs
/c/catalogs/{catalog_curie}
/c/catalogs/{catalog_curie}/resources
/c/catalogs/{catalog_curie}/resources/{resource_curie}
This change was made to simplify frontend handling of data returned by Catprez.
The catprez profile has been updated accordingly - Catalog items now display only a sample of their resources as part of the Catalog description - a full resource list can be viewed at the `/c/catalogs/{catalog_curie}/resources` endpoint

### DIFF
--- a/dev/dev-setup.py
+++ b/dev/dev-setup.py
@@ -137,6 +137,14 @@ def setup():
             ),
         ),
         (
+            "myfile17",
+            (
+                "agents.ttl",
+                open("tests/data/catprez/input/pd_democat.ttl", "rb"),
+                "application/octet-stream",
+            ),
+        ),
+        (
             "myfile16",
             (
                 "dublin_core_terms.ttl",
--- a/prez/app.py
+++ b/prez/app.py
@@ -142,27 +142,6 @@ async def app_shutdown():
         await async_client.aclose()
 
 
-@app.get("/", summary="Home page", tags=["Prez"])
-async def index(request: Request):
-    """Returns the following information about the API"""
-    # TODO connegp on request. don't need profiles for this
-    from prez.cache import (
-        prez_system_graph,
-        tbox_cache,
-    )  # importing at module level will get the empty graph before it's populated
-
-    prez_system_graph.add(
-        (
-            URIRef(settings.system_uri),
-            PREZ.currentTBOXCacheSize,
-            Literal(len(tbox_cache)),
-        )
-    )
-    return await return_rdf(
-        prez_system_graph, mediatype="text/anot+turtle", profile_headers={}
-    )
-
-
 def _get_sparql_service_description(request, format):
     """Return an RDF description of PROMS' read only SPARQL endpoint in a requested format
     :param rdf_fmt: 'turtle', 'n3', 'xml', 'json-ld'
--- a/prez/models/profiles_item.py
+++ b/prez/models/profiles_item.py
@@ -5,6 +5,7 @@ from pydantic import BaseModel, root_validator
 from rdflib import URIRef, PROF, Namespace
 
 from prez.cache import profiles_graph_cache
+from prez.config import settings
 from prez.services.curie_functions import get_uri_for_curie_id, get_curie_id_for_uri
 from prez.services.model_methods import get_classes
 
@@ -38,9 +39,7 @@ class ProfileItem(BaseModel):
         r = profiles_graph_cache.query(q)
         if len(r.bindings) > 0:
             values["classes"] = frozenset([prof.get("class") for prof in r.bindings])
-
-        values["label"] = profiles_graph_cache.value(
-            URIRef(values["uri"]),
-            URIRef("http://www.w3.org/ns/dx/conneg/altr-ext#hasLabelPredicate"),
-        )
+        label = values.get("label")
+        if not label:
+            values["label"] = settings.label_predicates[0]
         return values
--- a/prez/models/search_method.py
+++ b/prez/models/search_method.py
@@ -24,7 +24,9 @@ class SearchMethod(BaseModel):
     def __hash__(self):
         return hash(self.uri)
 
-    def populate_query(self, term, limit, offset, focus_to_filter, filter_to_focus, predicates):
+    def populate_query(
+        self, term, limit, offset, focus_to_filter, filter_to_focus, predicates
+    ):
         self.populated_query = self.template_query.substitute(
             {
                 "TERM": term,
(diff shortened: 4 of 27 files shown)


edmondchuc:
Did an initial test of the PR with the new changes in prez-ui (from this PR https://github.com/RDFLib/prez-ui/pull/116) and it looks good!

I tested it with the ANU records, which have around 400+ in a catalog.

Only thing I'll comment on is the ordering of the items. Is there any ordering applied? Should we order it by label like we are doing with spaceprez?

recalcitrantsupplant:
> Did an initial test of the PR with the new changes in prez-ui (from this PR [RDFLib/prez-ui#116](https://github.com/RDFLib/prez-ui/pull/116)) and it looks good!
> 
> I tested it with the ANU records, which have around 400+ in a catalog.
> 
> Only thing I'll comment on is the ordering of the items. Is there any ordering applied? Should we order it by label like we are doing with spaceprez?

ordering, no - I had a quick look at the ordering code and it is going to be difficult to adjust, suggest we revisit this - did you mean ordering for vocprez and not spaceprez? (though the query is the same)

edmondchuc:
> ordering, no - I had a quick look at the ordering code and it is going to be difficult to adjust, suggest we revisit this - did you mean ordering for vocprez and not spaceprez? (though the query is the same)

Ok, I thought it's using the same listing query, that's why I thought the ordering would be the same, but it seems like there's no ordering being applied at all.

Do you mind just elaborating a bit on why it's difficult and what are some of the issues?

recalcitrantsupplant:
It is just using the listing query yes, back when you added the order by label it only looks at `focus_item.top_level_listing` - which covered the VocPrez use case (and might cover this one too) - but it should really be generalised - we'd just need to test it.
The code block is:
```
{{
SELECT ?top_level_item ?child_item
WHERE {{
    {f'{uri_or_tl_item} a <{focus_item.base_class}> .{chr(10)}' if focus_item.top_level_listing else generate_focus_to_x_predicates(uri_or_tl_item, focus_to_child, focus_to_parent)}\

{f'''
    OPTIONAL {{
        {f'{uri_or_tl_item} <{ordering_predicate}> ?label .' if focus_item.top_level_listing else ""}
    }}
''' if settings.order_lists_by_label else ""}
}}
{f'''
{'ORDER BY ASC(?label)' if ordering_predicate else "ORDER BY ?top_level_item"}
```
ordering_predicate is at present the first predicate supplied in the settings.label_predicates (skos:prefLabel)

but... all of this can be simplified with sh:target/sh:select I proposed - you'd just write whatever subquery you want, and it would be inserted here. So I'm reluctant to spend too much time getting this implementation right. 

**Do you have a demo instance up somewhere we can run this against and see if it's an easy fix?**

edmondchuc:
@recalcitrantsupplant - yup, I've been testing it with the ANU catalog here https://api.anu.dev.kurrawong.ai/sparql. It has one catalog at the moment at https://anu.dev.kurrawong.ai/c/catalogs and around 400 records within it. The 400 records won't load in CatPrez right now because it's timing out but the data is available in the above SPARQL endpoint.

edmondchuc:
> It is just using the listing query yes

I guess I was just wondering how the spaceprez container items are being ordered but not the catprez implementation. You can see an example of the spaceprez stuff in the GSQ catalog [here](https://gsq-prez.dev.kurrawong.ai/s/datasets/dtst:qldgeofeatures/collections/qldgftrs:DevonianCarboniferousBasinTypes/items).

recalcitrantsupplant:
This is the current query that gets generated for Spaceprez Datasets
```
PREFIX dcterms: <http://purl.org/dc/terms/>
PREFIX prez: <https://prez.dev/>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>

CONSTRUCT {
  ?top_level_item a <http://www.w3.org/ns/dcat#Dataset> .
}
WHERE {
  ?top_level_item a <http://www.w3.org/ns/dcat#Dataset> .
  {
    SELECT ?top_level_item ?child_item
    WHERE {
      ?top_level_item a <http://www.w3.org/ns/dcat#Dataset> .
      OPTIONAL {
        ?top_level_item <http://www.w3.org/2004/02/skos/core#prefLabel> ?label .
      }
    }
    ORDER BY ASC(?label)
    LIMIT 20
    OFFSET 0
  }
}
```
And for Feature Collections, non top level, no ordering matches the conditionals in the query, but it may not be the desired behaviour..
```
PREFIX dcterms: <http://purl.org/dc/terms/>
PREFIX prez: <https://prez.dev/>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>

CONSTRUCT {
  <https://linked.data.gov.au/dataset/qldgeofeatures> ?focus_to_child ?child_item .
}
WHERE {
  <https://linked.data.gov.au/dataset/qldgeofeatures> ?focus_to_child ?child_item .
  VALUES ?focus_to_child { <http://www.w3.org/2000/01/rdf-schema#member> }
  {
    SELECT ?top_level_item ?child_item
    WHERE {
      <https://linked.data.gov.au/dataset/qldgeofeatures> ?focus_to_child ?child_item .
      VALUES ?focus_to_child { <http://www.w3.org/2000/01/rdf-schema#member> }
      OPTIONAL {
      }
    }
    ORDER BY ASC(?label)
    LIMIT 20
    OFFSET 0
  }
}
```

PR #174, opened 2023-11-20
David/use dependency injection

recalcitrantsupplant:
Looks like I merged the previous PR's in the wrong order - this is for previously reviewed/approved PR with dependency injection.


edmondchuc (approved):
Approving as reviewed in https://github.com/RDFLib/prez/pull/168.

PR #185, opened 2023-11-29
Fix broken profiles

### LINKED ISSUE #184: Most profile endpoints fail in latest version
All except for two profiles (DCAT & OGC API) fail on profile pages - `/profiles/<profile_curie>` for the current Prez version.

Error is:
`AttributeError: 'list' object has no attribute 'name'`

Tested against the IDN SPARQL endpoint.

recalcitrantsupplant:
Fixes profiles by creating a system store with Pyoxigraph for system data. At this point only profiles are added to this system graph. When a profiles related endpoint is called, this internal system repo is used. The RDFLib graph cache is still used elsewhere. Ultimately the Pyoxigraph store for system data can replace the various RDFLib caches.

Fixes #184

### DIFF
--- a/prez/app.py
+++ b/prez/app.py
@@ -14,6 +14,8 @@ from prez.dependencies import (
     get_pyoxi_store,
     load_local_data_to_oxigraph,
     get_oxrdflib_store,
+    get_system_store,
+    load_profile_data_to_oxigraph,
 )
 from prez.models.model_exceptions import (
     ClassNotFoundException,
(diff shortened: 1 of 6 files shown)


edmondchuc (approved):
Looks fine. Just curious why it wasn't working previously?

recalcitrantsupplant:
The profiles endpoints used the same repo dependency as other endpoints so profiles data was being looked for there rather than the RDFLib Graph where it was stored. And the tests didn't cover these endpoints.

jamiefeiss (approved):
Looks good. Would be good to know what the issue was.

Also we need _way_ more tests in Prez to cover every endpoint - i.e. multiple tests for both positive & negative tests using good and bad test data.

PR #188, opened 2023-12-11
Fix VocPrez queries not working on GraphDB

edmondchuc:
It looks like the SPARQL queries that are currently used for VocPrez to retrieve the top concepts and narrower concepts do not work in GraphDB.

I've re-written the queries to ensure that they now work with both Fuseki and GraphDB.

This bug was found by Engineers Australia where they were setting up an instance of VocPrez. Inform them once this fix has made it into a release.

### DIFF
--- a/prez/queries/vocprez.py
+++ b/prez/queries/vocprez.py
@@ -50,8 +50,6 @@ def get_concept_scheme_query(iri: str, bnode_depth: int) -> str:
     return dedent(query)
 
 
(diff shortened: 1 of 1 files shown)


recalcitrantsupplant:
Discussed in person - approved, will not test thoroughly as changes are subsumed in new version w/ rdframe.

recalcitrantsupplant (approved):

Issue #178, opened 2023-11-22
Prez link generation is slow


edmondchuc:
Currently, the `/object` endpoint takes an IRI and loads an object. It's slow because it does an N+1 database query for each IRI it finds within the resource to generate the prez links. So for a concept scheme with hundreds of top concepts, it will query the database for each one and attempt to generate a prez link. As you have mentioned in the past, this is the naive solution because it's slow but correct. 

Is there any way where we can use some of the info in the profiles to explicitly ignore relationships we don't want included during prez link generation? This kind of handling is similar to the custom queries we have now to handle the VocPrez endpoints for the progressive loading of large vocabs because we needed to ensure we were only loading the proximate information that's needed to render the UI and not a description of the entire vocab (like a SPARQL DESCRIBE query would).

I'm raising this because in the BGS case, they have very large vocabularies with hundreds or thousands of concepts and a render on the /object endpoint is extremely slow. For example, https://data-uat.bgs.ac.uk/object?uri=http://data.bgs.ac.uk/ref/Geochronology. If we can somehow instruct Prez to not include certain properties on routes like skos:hasTopConcept/skos:topConceptOf, then the loading of this vocab on the /object endpoint will be very fast, I think.

edmondchuc:
Related issue: https://github.com/RDFLib/prez/issues/166

hjohns:
Expect when caching is introduced, it may help. For large vocabs, the initial request will still have issues. Conditional prez-link generation may be needed. Issue potentially where there is a large number of top concepts. Further investigation required.

hjohns:
Solved in the current set of changes by using a simple cache, however the initial request load time is not improved.

Other options to discuss:
- Can we perform conditional link generation (Co-design session needed)
- Warming requests on startup

Meeting to be scheduled for next Monday.

edmondchuc:
My current understanding:

- `/object` requests are slow because regardless of what the class type is or what profile it uses, it always describes the entire object and processes the IRIs found to prez links.
- Slow generic object query via the normal system paths (`/s`, `/c`)
  - This is slow because the generic query gets the entire description of the object and processes the IRIs found to prez links
  - This does not affect large vocabularies because we are currently using a custom query which intentionally avoids retrieving relationships such as `skos:hasTopConcept` when rendering a vocab
[...]

PR #189, opened 2023-12-15
Improve performance for link generation

recalcitrantsupplant:
Improves the performance of link generation by using pyoxigraph for profiles queries instead of rdflib.

### DIFF
--- a/prez/app.py
+++ b/prez/app.py
@@ -15,7 +15,7 @@ from prez.dependencies import (
     load_local_data_to_oxigraph,
     get_oxrdflib_store,
     get_system_store,
-    load_profile_data_to_oxigraph,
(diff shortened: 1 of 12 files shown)


jamiefeiss (approved):
Looks good. Tested locally and it feels much faster now, thanks @recalcitrantsupplant

edmondchuc:
For context, as discussed in this issue https://github.com/RDFLib/prez/issues/178, this is an interim solution to speed up Prez endpoints that use the link generation function. The full solution will rely on the new RDFrame backend processing where we will be able to filter out certain link generations for certain endpoints.

@jamiefeiss, if you have tested it and there's an improvement, then great! Let's merge it in.

edmondchuc (approved):

PR #191, opened 2024-02-05
fix for reg-status and voc-der-modes

lalewis1:
add reg status and vocab derivation modes to the reference data sets. and alter the vocprez default profiles so that prez returns annotated rdf properly.


recalcitrantsupplant:
Tested locally

PR #209, opened 2024-02-29
merge connegp-refactor into rdframe-integration

lalewis1:
remove dependency on external connegp library


lalewis1:
I removed the last calls to the connegp library and now the tests can actually run.

But now we get lots of file not found errors. I wasn't seeing this locally because when you run the tests using the pycharm context menu you don't get these file not found errors. but when I run `poetry run pytest tests` locally, I do get the same errors. So just something to be aware of.

I'll go through and fix the file not found errors to try and get more of these tests passing before we merge it.

lalewis1:
@recalcitrantsupplant I have a feeling that some of the tests are dependant on the contents of the (ignored) `rdf` folder in the project root.

I noticed the tests that were failing for me, were actually different to what was failing on your branch, and the ones that are failing seem to be all related to missing data in the system_repo.

For context. I had 5 tests failing locally, then I removed my `rdf` folder and now I have 8 tests failing.

PR #211, opened 2024-03-11
Bump poetry version on release

edmondchuc:
We no longer set the version via an environment variable. It is now bumped on release and read at runtime from `pyproject.toml`.

This brings it in line with how we are managing versioning in prez-ui.


recalcitrantsupplant (approved):
sounds good

PR #215, opened 2024-03-28
Remove specification of "sh:path skos:prefLabel" in profile such that…

recalcitrantsupplant:
… all properties for a ConceptScheme are returned when looking at a ConceptScheme

Please test locally.

Specifying sh:path implies a closed profile i.e. any properties not specified under sh:path are excluded. This PR removes the specification of skos:prefLabel, reverting the profile to be open, so if the property skos:prefLabel is present, it will still be returned.


lalewis1:
Tested locally and vocabularies are now displaying correctly

lalewis1 (approved):

PR #216, opened 2024-04-02
build: add PREZ_VERSION environment variable to Dockerfile

edmondchuc:
I was hoping that the prez version is read from `pyproject.toml` but it is not being set when the container image is built via GH Actions for some reason.

This PR sets the `PREZ_VERSION` environment variable in the Dockerfile itself. The prez application will use this as the fallback version if it cannot read from `pyproject.toml`.


recalcitrantsupplant (approved):
haven't tested locally - have used same pattern in other images happy that this works

PR #222, opened 2024-04-10
Add annotations from semantic-background

lalewis1:
all tests passing and tested with prezUI v4


recalcitrantsupplant (approved):
approved - only adding static data. Will fix tests and merge

I rebased this on top of my branch hence all of the other changes - forgot to push through to my branch first so all of my changes are now in this PR too

PR #223, opened 2024-04-21
Docker build and wheel dist build fixes

ashleysommer:
I found some inconsistencies and bugs while building Prez in a docker container for deployment on Azure cloud.

This is a collection of five small commits related to building the Prez bdist wheel, installing it in the Docker container, and building the Docker container.

[1](https://github.com/RDFLib/prez/commit/f4db5690511d046c44dd1ab79df5b06fda6dd5bc) Use arg/env variable for venv creation instead of hardcoded location. (It used the venv in other calls, just not in the venv creation command). Explicitly call the venv version of pip3 to install the wheel, so it is obvious the built prez module is being installed in the venv site_packages.

[2](https://github.com/RDFLib/prez/commit/70fcb45da9eeb9db0d90157d6c910a5a2bb0a7d7) Don't copy the whole source tree into the final image, it already has the built prez module and all dependencies installed in the venv, it just needs main.py and supporting files.

[3](https://github.com/RDFLib/prez/commit/e314dca341c949a1f0faf363cbf9be691eadf0b0) Never include dist files in the docker env, this also ensures no dist files are included inside the sdist or wheel packages (poetry excludes packaging everything in gitignore).

[4](https://github.com/RDFLib/prez/commit/e026403259d5d5f8324793c8e7c4b3a435bfcc12) Don't put Dockerfile inside its own Docker build context. Its not needed for the container built, and this allows incremental builds to work properly.

[5](https://github.com/RDFLib/prez/commit/5128d684f19bf59690aa542ac056e6c6eda94cd1) Put prez module's pyproject.toml _inside_ the built prez module in the wheel package, this allows the config builder to find it when prez is installed as a system module. Looks like this is what #216 was trying to solve.
Modify the config builder to search inside the prez module for pyproject.toml if there is not one in the project top directory (enables the above functionality).
Also explicitly list all the files to include in a sdist, because should be different than the wheel bdist.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,5 @@ venv/
 .git/
 test_*.py
 .github/
+Dockerfile
+*.Dockerfile
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,6 @@ __pycache__/
 .idea/
 .pytest_cache/
 .env*
+dist/
 !.env-template
 rdf/
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,8 +28,8 @@ WORKDIR /app
 COPY . .
 
 RUN poetry build
-RUN python -m venv --system-site-packages /opt/venv
-RUN pip install --no-cache-dir dist/*.whl
+RUN python3 -m venv --system-site-packages ${VIRTUAL_ENV}
+RUN ${VIRTUAL_ENV}/bin/pip3 install --no-cache-dir dist/*.whl
 
 #
 # Final
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,7 @@ RUN apk update && \
       bash
 
 WORKDIR /app
-COPY . .
+# prez module is already built as a package and installed in $VIRTUAL_ENV as a library
+COPY main.py pyproject.toml ./
 
 ENTRYPOINT uvicorn prez.app:app --host=${HOST:-0.0.0.0} --port=${PORT:-8000} --proxy-headers
--- a/prez/config.py
+++ b/prez/config.py
@@ -67,9 +67,24 @@ class Settings(BaseSettings):
         values["prez_version"] = version
 
         if version is None or version == "":
-            values["prez_version"] = toml.load(
-                Path(Path(__file__).parent.parent) / "pyproject.toml"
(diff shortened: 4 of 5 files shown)


edmondchuc (approved):
Awesome, thanks for that Ashley.

Nice, I didn't know `pyproject.toml` had to be explicitly included like that. It's probably why the last few releases had the version missing since it couldn't find the file.

lalewis1:
@ashleysommer and @edmondchuc. Hey guys, so I just pulled these changes and when I try to run poetry install Poetry is complaining that:

The Poetry configuration is invalid:
  - data.packages[1] must not contain {'to'} properties

Just checking if this is something that is happening to you both as well or if it is just me?

Python version == 3.11
Poetry version  == 1.7.1

ashleysommer:
Hi @lalewis1 
I just checked the documentation, and I didn't realize that is a feature that was recently introduced in Poetry v1.8.0
https://python-poetry.org/docs/

More specifically, I found its a feature of `poetry-core` v1.9.0+

It's not in the docs for Poetry v1.7, so it must've been added after that. I thought that was a basic packaging feature that has been around for years.

Do you guys have a policy on supported Poetry version? When building Prez in a docker container its not an issue because latest Poetry is always used inside the container. But is it reasonable to expect developers on the code to always have the latest Poetry version, or should we maintain some kind of backwards compatibility?

lalewis1:
I think it would be fine to have the dependency on the latest version. It would seem there is no way to add the requirement in pyproject.toml (https://github.com/python-poetry/poetry/issues/6088), so perhaps the best thing will be to add a note to the readme.

lalewis1:
I just upgrade poetry to 1.8.2 and poetry-core 1.9.0 and it is now working as expected for me as well.

Thanks for the quick investigation!

ashleysommer:
Actually, you can specify the required poetry-core version used for the build backend here: https://github.com/RDFLib/prez/blob/0437c75a4c7e3620d38721f8d48f1087c86e073f/pyproject.toml#L47
Perhaps its worth bumping that to v1.9 now, then using that as the baseline minimum-supported poetry-core.

lalewis1:
yes good idea. I'll do it now.

PR #225, opened 2024-04-21
update poetry-core version to 1.9.0

lalewis1:
Dependency introduced by PR #223


ashleysommer:
Does this require a rebuild of poetry.lock? Sometimes poetry complains if you change a version in the toml file but don't update the lockfile.

ashleysommer:
I just double checked, it is fine, it doesn't need the poetry.lock file rebuilt.

PR #226, opened 2024-04-21
override anotated mediatypes

lalewis1:
Fixes missing filetype extension for annotated mediatype downloads.

see: http://localhost:8000/v/vocab?_profile=prfl:mem&_mediatype=text/anot+turtle


recalcitrantsupplant (approved):
one liner - looks fine tests pass

PR #230, opened 2024-06-03
Azure function

ashleysommer:
Not sure if this should be merged into Main or kept in a feature branch or just moved to docs.
This contains the changes needed to turn prez into an Azure function app.
It is based on the Azure Function App Python Runtime Container, the build is similar to the regular Prez Docker build, but with a couple of extra steps to make the installation directory and virtualenv directory compatible with Function Apps Python Runtime.


recalcitrantsupplant:
Discussed with Ashley that we will merge this to main as the Azure function code/files are in addition to Prez's "regular" files so do not otherwise impact running prez directly via Python or in the uvicorn Docker image.

recalcitrantsupplant (approved):
Approved as per comment

PR #233, opened 2024-06-17
Fix typo in connegp ontology prefix

ashleysommer:
Replace typo across all instances in the codebase


ashleysommer:
@nicholascar  in the [Spec](https://www.w3.org/TR/dx-prof-conneg/) `altr` namespace is `<http://www.w3.org/ns/dx/connegp/altr#>`
but under the [extensions section](https://www.w3.org/TR/dx-prof-conneg/#altr-extensions) the `altr-ext` namespace is only ever referred to with `<http://example.org/altr-ext#>`

It is reasonable that `altr-ext` should be `<http://www.w3.org/ns/dx/connegp/altr-ext#>` ?

PR #238, opened 2024-06-28
Specify Python 3.12.3

recalcitrantsupplant:
Specify Python 3.12.3; 3.12 currently defaults to 3.12.4 which fails on `pipx install poetry`


ashleysommer (approved):
I'll make a note here, this affects only the Python version in the built Docker container. It doesn't force a specific docker version in the pyproject.toml file, and doesn't affect running with other python versions locally.

PR #239, opened 2024-07-01
Add a POST endpoint for the /sparql passthru feature.

ashleysommer:
When Prez is behind Azure APIM, all URL query args are limited 2048 characters long.
sparql queries should not really be sent with GET requests anyway.
Add a POST endpoint for the /sparql passthru feature.
This causes the proxied query to the graph db to be called as a POST too


recalcitrantsupplant (approved):
Run/tested locally w/ pyoxigraph + Fuseki

PR #241, opened 2024-07-15
add missing labels for prez ontology

lalewis1:


recalcitrantsupplant:
The URI labels added are required by v3 but not v4 so no need to merge to main; could you check in settings if we can set branch protection on v3-maintenance as well i.e. cannot merge w/o pull request?

lalewis1:
ah, my apologies.

recalcitrantsupplant:
No worries I've added it now

PR #244, opened 2024-07-22
lint and format

lalewis1:
lint and format codebase with ruff / black.

I was just wanting to test out ruff last night. and got a bit carried away.

Take it or leave it. but all the tests are still passing and I did a test run with some data and prez ui and navigating around and everything is still working.


recalcitrantsupplant:
That's fine - do you want to update the versions in the pre commit hook/add ruff to it? https://github.com/RDFLib/prez/blob/main/.pre-commit-config.yaml . I was using this but need to reinstall it

PR #248, opened 2024-07-29
compute better filenames for returned rdf

lalewis1:
As per issue https://github.com/idn-au/catalogue-data/issues/35 The
returned RDF files did not have appropriate filenames.

This PR adds a filename where possible to the returned RDF via the `Content-Disposition` header.


recalcitrantsupplant:
Looks fine - could you add minimal tests to check the header comes back as expected for an object + listing

lalewis1:
Minimal tests added.

PR #252, opened 2024-08-15
fix: Search counts behave the same as listing counts

recalcitrantsupplant:
Search result counts are now consistent with listing counts:

If limit of 20 is requested:
- a search with limit N+1 = 21 is performed
  - if the results count = 21, then ">20" is returned,
  - otherwise the result count is returned as a string (e.g. 14 results returns "14")

Separately, the default search limit has been set to 10, and is configurable via "search_count_limit" on the app settings.


lalewis1 (approved):
yep all looks good.

PR #253, opened 2024-08-15
fix: add missing ontologies to annotations

recalcitrantsupplant:
Context_ontologies folder is not being used by any code, only annotations, so it has been removed

Some of the ontologies that were missing from annotations have been copied over from context_ontologies.
These can be identified as they are all *.nq files. In future they can be imported from labelify also once added there.


lalewis1 (approved):
looks fine thanks.

PR #255, opened 2024-08-16
Cannot use bare slash "/" as root_path in a Fast-API app.

ashleysommer:


recalcitrantsupplant (approved):
tested locally and in azure

PR #258, opened 2024-08-27
Add pure OGC Features API as option

recalcitrantsupplant:
This PR allows Prez to be run with endpoints in an "OGC Features" style i.e. `/collections/{collectionId}/items/{itemId}`, which looks for geo:FeatureCollections at the first level of the hierarchy, and geo:Features at the second.

To prevent the proliferation of internal links and similar profiles, it is intended/set up such that either but not both the current "extended" OGC Records/Features style API (`/catalogs/{catalogId} /collections/{collectionId}/items/{itemId}` **OR** the OGC Features style API as in this PR may be selected - not both. All other endpoints (CQL, Search, Profiles etc. remain the same across both configuration options).

This PR does not make material changes towards complete compliance with the spec. A follow up PR will add bbox filtering via a query param (NB CQL already allows a greater range of spatial filtering/predicates).

A small change was made to prioritise user specified prefixes in standard.ttl over those obtained from vann.

To utilise locally the easiest way to test is:
```
LOCAL_RDF_DIR=test_data
LOG_LEVEL=DEBUG
SPARQL_REPO_TYPE=pyoxigraph
API_FLAVOUR=ogc_features
```
To test with your own data, put it in a directory and change "test_data" to the location.


recalcitrantsupplant:
Most of this is rearranging splitting files and simple logic to include/exclude files etc.
I inadvertently included a small bugfix for query generation from SHACL - it enumerates the property shapes and includes the number in the node variable (i.e. ?node_n_...) such that the variable names don't collide.

ashleysommer:
While looking through this for review, I think I realized a misunderstanding I had about the router implementation in Prez4.
I thought the OGC-Features-API router was a secondary router bolted onto Prez as an optional extra to support GeoSpatial features (like the /g/ route on Prez v3). Afterall, the OCG-Features API spec is designed for querying geospatial features.
But I see now that serving datasets and vocabularies is also using this OGC Router in Prez v4. So configuring the router to run in "Pure" OGC-Features-API mode breaks the 3-level routing for serving vocabularies.

I thought the whole point of Prez was you could easily configure and add new API endpoints to suit the application, and serve whatever shapes of data you like based on the provided SHACL shapes. These changes in this PR look very hardcoded, and your comment about only allowing one router at a time leaves me very confused.

ashleysommer:
I believe what has happened at some point is a conflation of the OGC-Records API and the addition of an OGC-Features API.
The main feature shipped in Prez v4 was moving to the modern OGC-Records API, which was actually an new extended 3-layer OGC records API, for serving Vocabs, Catalogs, and Datasets.

My understanding was an OGC-Features API was being developed after the release of Prez v4, to be added alongside that for use in the BDR for serving geospatial features. But looking further into the specs for the OGC-Records API, I see it is designed to be an extension based off the Features API, and the routes are designed to be mostly the same, so it makes sense to combine them. But the OGC spec does not allow them to be literally the same endpoints (one serves Catalog metadata, dataset metadata, the other serves Geospatial features). 

What was implemented was neither. There is no `Capabilites` endpoint, and no `/conformance` endpoint, so it is neither a OGC-Features API nor an OGC-Records API, the prez implementation is an OGC-Like route pattern, with CQL support.

PR #261, opened 2024-08-27
Update Azure Function app files

ashleysommer:
There are several changes in this PR:
1) Remove requirements.txt file, because it is not really needed in the Azure-Container build, and the Azure-Function build generates its own requirements.txt
2) Move all azure-related files to ./azure/ directory, this cleans up the project root, as more files kept being added
3) Add an azure `publish.sh` script that will automatically build the required directory azure functiona-app structure and generate the requirements.txt file when needed, then execute the FUNC_CLI publish operation.
4) Adds a README.md in the /azure/ directory with more details.
5) A new file `patched_asgi_function_wrapper.py` that implements some basic patches on the `ASGIFunctionApp` wrapper and middleware implementation from Azure.Func module, that fixes a routing bug and fixes compatibility with ASGI Spec regarding enabling HTTPRequest States. (see the patch file for links to respective issue discussions).

Note, running the function app locally with the `func start` still works in this layout simply `cd ./azure` and `func start`, the `function_app.py` will detect its running in a subdirectory and will add `prez` from the parent directory into the Python Path, so the app will work as expected.


ashleysommer:
@recalcitrantsupplant  Your change looks fine, but you'll need to update the corresponding section in the `azure/README.md` file, it mentions executing the `publish.sh` script (it was change to publish_or_start), you'll need to document passing in the Action for publish vs start, and it also originally said that the FunctionAppName needs to be the _last_ argument (because that's how I implemented the args in the script) but your change means the FunctionAppName argument is now the 2nd argument (first arg after the Action).

recalcitrantsupplant:
Thanks have updated.

azure/README.md
```diff
@@ -1,17 +1,17 @@
...
 can be used to build a container image for deploying the app as an Azure Container App.
 
 ## Publishing
-There is a publish.sh script that can be used to publish the app to Azure.
+There is a publish_or_start.sh script that can be used to either build and run the function app locally, or publish the app to Azure.
 To call it, make sure you are not in the "azure" directory, instead run the script from the root of the project.
 
 ```bash
-./azure/publish.sh --extra-options <function-app-name>
+./azure/publish_or_start.sh start|publish --extra-options <function-app-name>
```
ashleysommer, line 11: Change \<function-app-name\> to come after start|publish, but _before_ --extra options

recalcitrantsupplant: Updated in [2857b44](https://github.com/RDFLib/prez/pull/261/commits/2857b446056dcb816ebb462938db1ac0964afd5a)

recalcitrantsupplant (approved):
Tested locally

ashleysommer:
Yeah, I tested it last night too, and tested it again after your changes. Ready to merge.

PR #268, opened 2024-09-29
David/shacl specified queryables

recalcitrantsupplant:
Allows OGC Features API Queryables (i.e. properties that Features can be filtered on) to be specified by SHACL.

This enables property path expressions in the graph to be "pre defined" and queried as a "flat" property (queryable). Without this functionality, only direct properties are available as Queryables, either limiting the scope of what can be queried, or requiring flattening of the graph.

This requires some manual steps to:
- Define the queryable propery paths in the graph
- Optionally, specify additional metadata about the queryables (enumerations, datatypes etc.)

A full readme is provided in README-OGC-Features.md


lalewis1 (approved):
nice work again. I couldn't catch any errors.

PR #273, opened 2024-10-04
merge lawson/label_predicates into maintenance-v3

lalewis1:
small fix to allow setting of labelling predicates via environment variable.

### DIFF
--- a/prez/config.py
+++ b/prez/config.py
@@ -1,10 +1,10 @@
 from os import environ
 from pathlib import Path
-from typing import Optional, Union, Any, Dict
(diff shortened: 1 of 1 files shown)


recalcitrantsupplant (changes requested):
Looks fine but can you change root_validator to field_validator

https://docs.pydantic.dev/latest/migration/#validator-and-root_validator-are-deprecated

lalewis1:
thanks good catch

recalcitrantsupplant (approved):
good catch on the catch

PR #276, opened 2024-10-08
lawson/predicate validation

lalewis1:
- **fix: allow overriding predicates via env var**
- **fix: bump python version**
- **chore: add test cases**


recalcitrantsupplant (approved):
Looks good

PR #277, opened 2024-10-11
Remote profiles bugfix and update to Catalog hierarchy

recalcitrantsupplant:
Updates Catalog hierarchy to dcat:Catalog / dcat:Resource. Retains dcat:Catalog / dcat:Dataset which is the "parent" of geo:FeatureCollections in "spaceprez".
Bugfix for remote profiles not being loaded to the system (pyoxigraph) store.


lalewis1 (approved):
lgtm

PR #278, opened 2024-10-13
David/ogc features bugfixes

recalcitrantsupplant:
CQL property value filtering assumes URI if starts with http/https.

Add dynamic mount path for OGC Features API; including dynamically modifying hierarchy level of OGC Features endpoints based on where it is mounted.

The latter is required for the MERIT API in BDR such that it can be mounted at e.g. /features


lalewis1 (approved):
lgtm

tested locally with default configuration and it works fine.

PR #279, opened 2024-10-14
HTML redirect to Prez UI

avillar:
Redirect object requests to their HTML views when either `text/html` or no media type is requested.


recalcitrantsupplant (approved):
Looks good. Only comment would be here where there's >1 link, I don't think we have a good solution. We could send someone to the /object page in that scenario?

```
prez_link = item_graph.value(subject=item_uri, predicate=URIRef("https://prez.dev/link"), any=True)
```

I discussed with Howard and we plan to open a separate discussion in the repo here, the main points were:
- URL redirects from the frontend would be good but would need a web server unless it's javascript
- we have customers that only want to expose the API to the frontend, so the solution here wouldn't work I believe - perhaps in those cases though connegp is less of a concern.

PR #280, opened 2024-10-15
Deprioritise Alternates Profile

recalcitrantsupplant:
You will now only get the alternates profile if you specifically request it. Otherwise if any other profile is available, you'll get that. This will prevent unintended behaviour in the UI where the Alternates Profile was equally "ranked" with another profile, so either could be returned on page refresh.


recalcitrantsupplant:
@hjohns fyi

lalewis1 (approved):
looks good

PR #281, opened 2024-10-16
Only retrieve remote Prez profiles; count query bug; annotation performance

recalcitrantsupplant:
1. Ensure only prez related profiles are loaded from remote stores by updating the profile selection query
2. Type profiles used to declare default profiles as prez:IndexProfiles. This makes 1 easier
3. Fix count query bug. Counts could previously be wrong in some instances - moved DISTINCT to correct place.
4. Improve performance of annotations retrieval. Identified slow code and changed to a faster implementation.


recalcitrantsupplant:
@avillar I can't add you as a reviewer in this repo, will see if we can add you - but do let me know if this works OK for you

lalewis1 (approved):
looks good, counts are returning correctly now.

PR #284, opened 2024-10-17
OGC Features queryables bugfix + cleaner curie generation

recalcitrantsupplant:
1. OGC Features bugfix for auto generated queryables.
2. Add upfront validation of curie -> URI and raise URI not found exception if appropriate. 
3. Add special characters to those excluded from CURIE generation: "aeiou!@#$%^&*()_+-=,."

### DIFF
--- a/prez/dependencies.py
+++ b/prez/dependencies.py
@@ -24,7 +24,7 @@ from prez.enums import (
     JSONMediaType,
     GeoJSONMediaType,
 )
-from prez.exceptions.model_exceptions import NoEndpointNodeshapeException
+from prez.exceptions.model_exceptions import NoEndpointNodeshapeException, URINotFoundException
 from prez.models.query_params import QueryParams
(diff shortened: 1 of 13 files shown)


lalewis1:
Looks good, seems to work fine.

However, I am still getting a parser issue in prez-ui, even
without any prefixes with '.'s in them.

Will add to the relevant issue thread.

probably safer not to have any of those characters in the prefixes anyway.

lalewis1 (approved):
thanks. looks good

recalcitrantsupplant:
good catch, yes I think you're right. We'll probably end up with the ns[1-9] type auto generated prefixes from RDFLib in a lot of cases "2.0" -> "" -> ns1 etc.

lalewis1:
@recalcitrantsupplant , I have implemented the prefix fix. you'll see that i am hashing the uri to get a prefix. I wonder if we could also do this in place of 'ns1' 'ns2' style prefixes? did you say that was an rdflib thing? or is that something we can control from Prez?

lalewis1 (approved):

PR #288, opened 2024-10-24
Do not attempt to create link for certain classes.

### LINKED ISSUE #286: Object Listing function Bug
Found issue in https://github.com/RDFLib/prez/blob/94fe37d6c6883d845a0794365de215b728bc0b09/prez/services/link_generation.py#L211-L224

whereby it is possible that an invalid sparql query can be generated. of the form

```sparql
select
where {
    <ex:1> <ex:2> <ex:3> .
}
```

i.e. nothing in the select clause which causes an error.

recalcitrantsupplant:
fixes #286


lalewis1 (approved):
brilliant thanks.

tests passing locally and 286 is resolved with the anu catalogue data too.

Issue #287, opened 2024-10-24
Make vowel removal optional


KoalaGeo:
Would it be possible to make the vowel removal in the prefix generation optional, using a `remove_vowels` variable (default = true) which should be set in the config? 

I believe this could be acheived like:
```
def generate_new_prefix(uri, remove_vowels=True):
    """
    Generates a new prefix for a uri
    """
    parsed_url = urlparse(uri)
    if bool(parsed_url.fragment):
        ns = f"{parsed_url.scheme}://{parsed_url.netloc}{parsed_url.path}#"
    else:
        ns = f'{parsed_url.scheme}://{parsed_url.netloc}{parsed_url.path.rsplit("/", 1)[0]}/'

    split_prefix_path = ns[:-1].rsplit("/", 1)
    if len(split_prefix_path) > 1:
        to_generate_prefix_from = split_prefix_path[-1].lower()
        # attempt to just use the last part of the path prior to the fragment or "identifier"
        if len(to_generate_prefix_from) <= 6:
            proposed_prefix = to_generate_prefix_from
            if not prefix_registered(proposed_prefix):
                prefix_graph.bind(proposed_prefix, ns)
                return

        if remove_vowels:
            # remove vowels to reduce length
            proposed_prefix = "".join(
                [c for c in to_generate_prefix_from if c not in "aeiou!@#$%^&*()_+-=,."]
            )
            if not valid_prefix(proposed_prefix):
                # if we still can't get a nice prefix, use an ugly but valid one using a hash of the IRI
                proposed_prefix = f"ns{hash(to_generate_prefix_from)}"
            if not prefix_registered(proposed_prefix):
                prefix_graph.bind(proposed_prefix, ns)
                return
        else:
            # Use the original string without removing vowels
            proposed_prefix = to_generate_prefix_from
            if not valid_prefix(proposed_prefix):
                # if we still can't get a nice prefix, use an ugly but valid one using a hash of the IRI
                proposed_prefix = f"ns{hash(to_generate_prefix_from)}"
            if not prefix_registered(proposed_prefix):
                prefix_graph.bind(proposed_prefix, ns)
                return
    else:
        raise ValueError("Couldn't generate a prefix for the URI")
```

recalcitrantsupplant:
Hi Edd, I'm happy to remove it altogether rather than having another configuration option - would that suit? Thoughts @lalewis1 ?

[...]

PR #289, opened 2024-10-25
keep vowels in generated prefixes #287

lalewis1:
keep vowels but remove all punctuation characters as per #287


recalcitrantsupplant (approved):
Looks good thanks

PR #291, opened 2024-10-28
Root path (and reverse proxy) fixes

avillar:


avillar:
A few tests are failing, which I'm not surprised about. Looks like they're for OGC API Features stuff.

recalcitrantsupplant:
Thanks Alejandro I'll try get to this soon

recalcitrantsupplant (approved):
I've set proxy headers to False by default - aware we previously had it set to True, so you'll need to explicitly enable it. Have set the default forwarded_allow_ips to 127.0.0.1 in both the Dockerfile and config.py/settings.

I don't usually touch these settings - let me know if any concerns re. the above, if not will merge tomorrow.

avillar:
`--proxy-headers` needs to be a flag (just pushed a fix), but other than that I think it works :)

PR #294, opened 2024-11-01
David/check header

recalcitrantsupplant:
Checks headers for a specified header with a specified value, can be used for e.g. API Subscriptions.
Updates Pyoxigraph to fix unrelated test failure with OGC Features API - Lawson is separately investigating.


lalewis1 (approved):
looks good to me.
tested locally and it does what it says on the box.
tests passing, but I will investigate further.

PR #295, opened 2024-11-05
Ensure for SPARQL passthrough queries headers are passed through from…

recalcitrantsupplant:
… the original request and not the defaults supplied by httpx AsyncClient (e.g. compression).

Return more sensible error messages.

Tried to test this yesterday but couldn't replicate the issue with the pytest testclient. Suggest merging for now and this may be easier to add tests for once the settings are in a dependency.


lalewis1 (approved):
looks good to me

PR #298, opened 2024-11-07
David/dynamic ogc features

recalcitrantsupplant:
- Allow remote specification of OGC Features API endpoint definitions.
- OGC Features bugfixes
- temporarily disable arm64 linux dockerfile builds


lalewis1 (approved):
thanks looks good

PR #299, opened 2024-11-08
David/ogc features fixes

recalcitrantsupplant:
feat: introduce Features Router to support consistent redirection behaviour i.e. "/" redirects to "" on the end of routes. Fix count bug where count is zero.


recalcitrantsupplant:
Have realised this creates a bug where /items on the non Features route doesn't work now.

PR #301, opened 2024-11-11
feat: add "subscription-key" as query parameter, to support Azure and…

recalcitrantsupplant:
… other APIs that wish to use API keys.


lalewis1 (approved):
looks good thanks

PR #305, opened 2024-11-14
Lawson/fts

recalcitrantsupplant:
Creates a search query which is the union of:
- One full text search query for all direct predicates e.g. ?focus_node rdfs:label "search term"; and
- One full text search query PER shacl path defined as a `https://prez.dev/ont/JenaFTSPropertyShape`

Can run Fuseki FTS using `python dev/dev-setup.py`
Then query:
`/search?q=*yue*&predicates=real&predicates=https://schema.org/headline&predicates=http://www.w3.org/2000/01/rdf-schema%23label`
The inner select clause is of the form (for two direct predicates + one path):
```sparql
      {
        {
          (?focus_node ?weight ?match ?g ?pred ) <http://jena.apache.org/text#query> (<https://schema.org/headline> rdfs:label "*yue*" )
        }
        UNION
        {
          (?fts_search_node ?weight ?match ?g ?pred ) <http://jena.apache.org/text#query> (<https://schema.org/description> "*yue*" ) .
          ?prof_101_node_1 <http://www.w3.org/ns/sosa/hasResult> ?fts_search_node .
          ?focus_node <http://www.w3.org/ns/sosa/isFeatureOfInterestOf> ?prof_101_node_1
        }
      }
```
Todo: add variables used in property path to selected variables so they are available to the construct query.


lalewis1:
I've added blank node filtering such that if the focus node is a blank node then it will be removed from the search results.

I also dropped a few commits to remove the testcontainers data and dependency, which were causing some issues with git.

recalcitrantsupplant:
looks good, only just occurred to me, I think isIRI in this case would be equivalent to !isBlank and probably a little less code. Can update later

lalewis1 (approved):

PR #310, opened 2024-11-20
feat: allow prefixed env vars

lalewis1:
For label_predicates, description_predicates etc. this commit allows you
to set the values as prefixed values. for example

```
# .env
LABEL_PREDICATES='["schema:headline", "dcterms:title"]'
```

this should work for any of the default bound prefixes in rdflib

> [!WARNING]  
> i had to change the test cases becuase this now considers "1" to be an invalid prefix
> this may or may not be a breaking change actually, so please consider carefully.

### DIFF
--- a/prez/config.py
+++ b/prez/config.py
@@ -5,7 +5,7 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 import toml
 from pydantic import field_validator
 from pydantic_settings import BaseSettings
-from rdflib import DCTERMS, RDFS, SDO, URIRef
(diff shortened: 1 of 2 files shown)


recalcitrantsupplant:
I'd prefer not to do this as it relies on us knowing what namespace the prefix is mapped to, in common cases this may be clear but even for say schema: is that referring to the http or https version of schema, and for anything less common where would the mapping be kept, and how would you guarantee the prefix:namespace mappings are 1:1.

I realise perhaps counter intuitively the prez settings use the RDFLib Namespaces so can be written in the prefixed form

PR #311, opened 2024-11-21
ci: migrate from python-semantic-release to semantic-release

lalewis1:
Python semantic release was only used to auto update the version
variable in pyproject.toml, but this does not work with branch
protection rules and so is not neccessary anymore. Thus it is better to
use the more common semantic-release action which works with branch
protection rules as it does not need to commit any files as part of the
release.


recalcitrantsupplant (approved):
Looks good - I've updated the versions of the two actions. Lets run it and see if any issues.

PR #316, opened 2024-12-04
fix: remove use of JSON LD from CQL JSON parsing

recalcitrantsupplant:
Removes the use of JSON LD with CQL JSON to simplify parsing/processing of CQL JSON.

In order to do this, CQL JSON properties that begin with "http" are interpreted as URIs. This is a reasonable but not foolproof assumption that may need revision.

Tests that previously only passed with Fuseki now work with the inclusion of GeoSPARQL support in Oxigraph and have been uncommented.


lalewis1 (approved):
looks ok to me. admittedly, a bit hard to really understand all the changes. but it works locally for me and the tests are passing.

just needs to be rebased before merge.

PR #318, opened 2025-01-08
bugfix: Allow for empty Accept headers. Browsers/Curl/httpx do not by…

recalcitrantsupplant:
… default supply empty Accept headers, postman/hopscotch etc. do.


lalewis1 (approved):
looks good to me. good catch.

PR #319, opened 2025-01-08
feat: allow multiple template SPARQL queries for an object/item endpo…

recalcitrantsupplant:
…int. Remove unused file based template queries - configuration was complicated with having to specify file paths - and not currently used by anyone. Only data repo template queries are now supported.

From dev documentation, to add a template SPARQL query, use the form:

```sparql
PREFIX prez: <https://prez.dev/ont/>
INSERT DATA { GRAPH <https://prez/system> {
    [ a prez:TemplateQuery ;
        rdf:value """<template_queries>""" ;
        prez:forEndpoint "http://www.opengis.net/ogcapi-features-1/1.0/feature" ;
    ]
    }}
```

Prez will now handle multiple literal values for rdf:value and execute all of these queries, returning the graph merge of the result.


ashleysommer (approved):
looks great! I'm glad it wasn't too much code to change to enable this feature.

PR #320, opened 2025-01-10
David/geojson for cql items

recalcitrantsupplant:
This change introduces/fixes GeoJSON support on the /cql and /object endpoints. It reuses OGC Features code that generates GeoJSON. There are no major code changes in this change, it is mostly cleanup/refactoring etc.

NB a set of supporting rendering functions was moved from listings.py to renderer.py to make it easier to reuse them across the OGC Features endpoints and the regular endpoints.


lalewis1 (approved):
ok, looks alright to me.

PR #321, opened 2025-01-10
fix: github action creates docker images based on sem release

recalcitrantsupplant:


recalcitrantsupplant:
going to merge this as it's hard to test locally (have tried with `act`); need to push to main to trigger it

PR #330, opened 2025-01-20
fix: add/raise appropriate error for /cql endpoint where no filter is provided

recalcitrantsupplant:


lalewis1 (approved):
looks ok. thanks

PR #335, opened 2025-03-03
feat: add support for human readable GeoJSON and other changes

recalcitrantsupplant:
A number of changes, the main one being human readable GeoJSON. The others are small features/fixes/performance updates.

feature: add support for human readable GeoJSON. This is implemented by prez providing "annotated" subgraphs to the rdf2geojson dependency which flattens the graph, providing GeoJSON which has labels instead of IRIs, and flattens the view to support rendering in tools such as ArcGIS and QGIS.
feature: allow use of "shext:allPredicateValues" on sequence paths. This means that all properties of a non focus node can easily be returned. Example snippet:

```turtle
    sh:property [
        sh:path (
            sh:union (
                     shext:allPredicateValues
                     ( geo:hasGeometry geo:asWKT )
                     ( schema:temporal shext:allPredicateValues )
```
fix: add missing annotated mediatypes to profiles so they can be requested/returned
perf: request/retrieve ntriples from the triplestore by default from the triplestore. This slightly increases the data transfer amount (which is compressed by default), but parses faster in RDFLib so on the whole using ntriples is faster than turtle.
feature: allow root_path specification in fastapi, this was motivated to support deploying behind API Management in Azure, so that /docs renders correctly. This feature exposes the configuration correctly on the Prez/FastAPI side, however there appears to be a path calculation issue when deploying to Azure and using API Management with a non "root" path. Further investigation in Azure is required.


prez/app.py
```diff
@@ -28,8 +28,8 @@
     InvalidSPARQLQueryException,
     NoEndpointNodeshapeException,
     NoProfilesException,
-    PrefixNotFoundException,
-    URINotFoundException, MissingFilterQueryError,
```
lalewis1, line 32: MissingFilterQueryError should still be imported

recalcitrantsupplant: fixed now

lalewis1 (approved):
Yeah looks ok to me.
there are a few unused imports from `ruff check`
but tests pass and I can't see any major issues

PR #341, opened 2025-03-21
fix: reduce OGC Features link header length for collections

recalcitrantsupplant:
reduce header length in OGC Features API listings (at /features/collections/). Issue should not appear at .../items as the links here are not per item. Reduce number of mediatypes shown at /features/collections/ in order to reduce link header size. NB other RDF mediatypes, including annotated ones, can still be requested/returned using the Accept header or _mediatype query parameter as before.


lalewis1 (approved):
looks ok, tested locally and it's working

PR #342, opened 2025-03-24
perf: improve classes query

recalcitrantsupplant:
improve performance of classes SPARQL query by changing the values clause to an equivalent inline data clause. This also accommodates a bug in qlever.

### DIFF
--- a/prez/services/query_generation/classes.py
+++ b/prez/services/query_generation/classes.py
@@ -12,9 +12,10 @@ from sparql_grammar_pydantic import (
     SubSelect,
     TriplesBlock,
     TriplesSameSubjectPath,
-    ValuesClause,
     Var,
     WhereClause,
+    GraphPatternNotTriples,
+    InlineData,
 )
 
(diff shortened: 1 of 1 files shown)


lalewis1:
I think it is ok, just investigating another qlever bug to see if it is related.

it is still failing for search queries during link generation.

recalcitrantsupplant:
oh right, as in the generation and not when you try to go to the link? I actually did try search queries and it seemed to be working OK for me. I configured SEARCH_PREDICATES=["https://schema.org/location"], no other config

lalewis1:
yeah i think its to do with limits, using the koenigsnet data if i search for catalog there are like 20,000 results and then it tries to link generate for all of them and just gets completely stuck / blocked.

lalewis1:
i've identified the issue as being elsewhere.

tracking it here https://github.com/RDFLib/prez/issues/344

will approve this PR

lalewis1 (approved):
yep looks good

PR #343, opened 2025-03-25
fix: range of sh:path should be blank node not RDF Collection

recalcitrantsupplant:
Shacl path should have a bnode as it's range not an RDF Collection. This is a basic error/typo that has been present for some time and somehow not caused issues until now (potentially due to default RDF ordering in RDFLib). Using an RDF Collection meant that one of either the union of property paths OR the shacl:union property itself were processed, potentially RDFLib typically ordered these internally in such a way that the list and not sh:union was processed. Olis seems to have disrupted this. In any case, use of an RDF Collection is wrong.

The following two changes have been made:
1. Fix the processing of sh:union statements, these were the only thing specified off of RDF Collections from sh:path
2. Fix the relevant profiles to use blank nodes instead of RDF Collections ( "(" -> "[" )

### DIFF
--- a/prez/reference_data/profiles/ogc_features.ttl
+++ b/prez/reference_data/profiles/ogc_features.ttl
@@ -55,13 +55,13 @@ prez:OGCFeaturesMinimalProps a
         "text/turtle" ;
     sh:property
         [ sh:path
-                  (
+                  [
                       sh:union (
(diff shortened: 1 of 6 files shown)


lalewis1 (approved):
no issues, thanks, looks fine

PR #345, opened 2025-03-27
fix: order by now works correctly for listing endpoints

recalcitrantsupplant:
- order by now works correctly for listing endpoints, previously it only worked for hierarchical endpoints
- changes to align concept hierarchy ordering so it still works

specifying order by in the query string argument will:
- insert a triple into the subselect for focus nodes: `?focus_node <order by IRI> ?order_by_val`
- add a matching order by clause `ORDER BY ?order_by_val`

If you are using the default endpoints, with skos:ConceptScheme at the second hierarchy level, and wish to list all skosConceptScheme instances Prez can see _across multiple dcat:Catalogs, i.e. the vocabs are under different catalogues at the top level_, you can use the /cql endpoint to do this. You must:

1. Ensure the profile you are going to use constrains the pseudo class `prez:CQLFilterResult` - this means the profile can be used with instances of that class. With CQL, as there can be multiple classes returned, we use the placeholder `prez:CQLFilterResult`.
2. _Add whichever other properties you want displayed to the profile (this has nothing to do with ordering). You do not need to return the properties that you intend to order by (though it may be a good idea to do this in order to prevent confusion)_
Here is an example profile that constrains `prez:CQLFilterResult`:
https://github.com/RDFLib/prez/blob/200a60b5b31dc8b6443fd3b5b28c37f1577129a0/prez/reference_data/profiles/ogc_records_profile.ttl#L144

At runtime:
```
import httpx
import json

filter_obj = {
    "op": "=",
    "args": [
        {"property": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"},
        "http://www.w3.org/2004/02/skos/core#ConceptScheme"
    ]
}

params = {
    "filter": json.dumps(filter_obj),
    "_profile": "ogc-schemes-listing",
    "order_by": "http://www.w3.org/2004/02/skos/core#prefLabel",
    "_mediatype": "text/turtle"
}

response = httpx.get("http://localhost:8000/cql", params=params)
```


recalcitrantsupplant:
To filter to vocabs and a particular property e.g. a schema:keyword, example CQL JSON:

```json
{
  "op": "and",
  "args": [
    {
      "op": "=",
      "args": [
        {
          "property": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
        },
         "http://www.w3.org/2004/02/skos/core#ConceptScheme"
      ]
    },
    {
      "op": "=",
      "args": [
        {
          "property": "https://schema.org/"
        },
        "https://my-keyword"
      ]
    }
  ]
}
```

To filter on multiple keywords use "in" with an array.

recalcitrantsupplant:
@lalewis1 I know you requested this recently - was there data you were also able to test this with?

jamiefeiss:
Tested locally with IDN data. The CQL listing works well, though I haven't made a custom profile to select certain predicates. I'm not sure if I'm using the `order_by` query param correctly, but specifying `https://schema.org/name` seems to work, not `prez:label`, etc, but I'm not noticing any ordering:

Screenshot 2025-03-28 at 4 08 49 pm

Also I can't seem to get any results with search.

recalcitrantsupplant:
Screenshared w/ Jamie, langtags are ordered after non langtagged literals, so introduced string function i.e. `STR(?order_by_val)` in order condition in order to "ignore" langtags.
NB ordering with `prez:label` is not possible, these triples are calculated or derived as annotations intended for the UI, they are not materialised so the generated queries cannot use them.

jamiefeiss:
Did a quick test with just the API, looks like ordering works nicely. I'll need to update Prez UI to support the new ordering.

Search is still broken, I can't get any results. Any ideas?

recalcitrantsupplant:
Please try again. I was conflating the order by predicate and order by value, these are now separate things and only the relevant one is passed to the query constructor.
In summary:
1. order by query parameter sets the order by "predicate" e.g. `?focus_node <order_by_predicate> ?order_by_val`. `?order_by_val` is then defaulted to a var with that name
2. for search, `order_by_predicate` does not need to be set. It is used within within the (complicated) search queries, the output of which is the `order_by_val`, which in this case is bound to ?weight. The query constructor gets this from the search queries.

jamiefeiss:
Search works now, thanks. Does it order by weight by default? If I do ?limit=1 I don't get the result with the highest weight

recalcitrantsupplant:
Search orders by weight always - OK I'll check again. It does use STR( ) with the integer which I was concerned about but still seemed to work fine.

recalcitrantsupplant:
OK, so for other listing queries, prez extracts the subselect and runs this with a higher "listing" limit to get the count. 
This prevents retrieving all properties for all objects just to get a count of objects. So with default settings you get:
1. 10 objects and their properties
2. a count of the total number of objects (the actual number if less than the listing count limit which is 100 by default, otherwise ">100")
With the search query, the subselect is expensive, so running it even with limit 100 was often slow.
For this reason, is just run with limit + 1. Unfortunately this means you request 10 results and get 11 etc.
So your results should be ordered by weight, and it should include the top N results, so limit=1 ends up being limit=2. This is what I'm seeing when I tested just now, is it consistent with what you are getting? We could add more logic to remove the additional result, but I think it might be more efficient/easier to do in the FE.

jamiefeiss:
Yes I noticed that, I'll have to cater for the extra result in the frontend, though I'm not sure how to tell which result to omit (the limit+1 result).

The issue is that I know that the top result should have a weight of 130, but the 2 results I get for limit=1 are both of weight 30.

With no limit set with sorting implemented in the frontend:
Screenshot 2025-04-01 at 5 29 36 pm

With limit=1:
Screenshot 2025-04-01 at 5 30 12 pm

recalcitrantsupplant:
In terms of weight I guess they're both as valid as each other, and if someone hasn't found what they're after in the top N (say 10) they should probably add more filters/change the search term.
Perhaps in a follow up we can do a more comprehensive order by weight + label (or weight + something else).

recalcitrantsupplant:
The search now orders by weight without the STR function. Not sure why in my previous testing I thought it didn't affect things. Are you able to test again? I'll also add some further test cases.
- concept hierarchy and search now have default order by predicates/values that come out in the final query.
- the order by query param can be used for regular listing queries (non search/concept hierarchy), and it also overrides search and concept hierarchy defaults if specified. It adds an additional triple to the subselect and orders by the val that is bound in this triple.

lalewis1:
sorry, i can't figure out how to actually use this?

i'm trying to set submit /catalogs?order_by=label, but obviously that is not correct because I get a 500 error.

jamiefeiss:
You need to specify the predicate IRI, like `?order_by=https://schema.org/name`. You might need to URL encode the IRI, especially if it has a hash in it like skos

jamiefeiss:
@recalcitrantsupplant search works great now, thanks!

jamiefeiss (approved):

Issue #344, opened 2025-03-26
DRAFT: search query doesn't respect limit clause


lalewis1:
A change in commit ... has introduced a bug where the listing query no longer respects the limit clause.

...

recalcitrantsupplant:
I don't think this is still relevant/accurate - the properties do not have a limit restriction on them and are not intended to. I believe the other fixes we made in targeting prez:Search should have resolved the initial bug we had with Qlever.

PR #346, opened 2025-03-27
fix: exclude class triple for plain search and cql searches

recalcitrantsupplant:
Summary: This PR ensures that when a search or CQL query is run _directly from the /search or /CQL endpoints **only_**, there is no focus node class restriction triple added i.e. the `?focus_node rdf:type <class>`. It also ensures that search/cql queries are directed to appropriate profiles _in the default profiles_.

Previously the default profiles used did not add this triple. This change ensures there is no restriction on the profiles.

Implementation: add a check to the umbrella query builder that if a search or CQL query is received ONLY on the /search or /cql endpoints, then the class information for the types of objects is excluded from the generated SPARQL query. search and CQL queries received on other listing endpoints will still apply the class filter relevant to that endpoint.\

@lalewis1 this is the root cause of https://github.com/RDFLib/prez/issues/344 (the search queries shouldn't have been directed to that profile by default).

If someone requests a profile that displays all properties of objects that have thousands of outbound member links, and uses this with a search/cql (i.e. a listing) query , I think it's correct to return this data, though we wouldn't expect the performance to be good. We should just ensure that the default listing profiles do not do this (as per this fix).

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -136,7 +136,7 @@ prez:OGCRecordsProfile
     altr-ext:hasNodeShape [
         a sh:NodeShape ;
         sh:targetClass prof:Profile , dcat:Catalog , dcat:Resource , skos:Concept , geo:Feature , geo:FeatureCollection
-                               , skos:Collection , rdf:Resource , prez:SearchResult , prez:CQLObjectList ;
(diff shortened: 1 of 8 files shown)


lalewis1 (approved):
yeah this is great. works as expected for me.

PR #348, opened 2025-03-28
fix: don't assume RDF from remote endpoint is n-triples

lalewis1:
instead get the content-type from the response and fall back to
n-triples if not given


recalcitrantsupplant (approved):
sounnds good

PR #350, opened 2025-04-09
fix: add missing api docs for /object endpoint. change from uri to iri

recalcitrantsupplant:
adds correct api documentation for the `/object` endpoint.

NB docs point users to use `iri` however `uri` is still supported so this is not a breaking change.

image


recalcitrantsupplant:
I should add profile and mediatype too hold on a minute..

recalcitrantsupplant:
rebase so api docs for facet_profile are visible

hjohns (approved):
Looks good.

PR #353, opened 2025-04-17
feat: add faceting, path aliases, missing query params from object endpoints

recalcitrantsupplant:
Adds the following functionality:

1. Faceted search for categorical properties. 
Faceting profiles are no different to prez profiles for property selection. However for now they should define properties for faceting within a `sh:property [ sh:path [ sh:union ( ...` construct.

For example to facet on `rdf:type` a profile like this could be used:
```turtle
<https://prez.dev/profile/facet-by-type>
    dcterms:identifier "facet-type"^^xsd:token ;
    dcterms:title "Facet things by type" ;
    dcterms:description "Allows faceting by rdf:type" ;
    sh:property [ sh:path [ sh:union ( rdf:type ) ] ] .
```
The faceting can be triggered, currently only on listing endpoints (including /search & /cql) with the facet_profile query parameter, using the `dcterms:identifier`, iri, or curie of the profile. This generates a faceted search query using the subselect of the main focus node query prez generates and merges the results with the data graph. 

Example QSA:
`?facet_profile=my-faceting-profile`  
The faceting results are of the form:
```turtle
[] prez:facetCount 1 ;
    prez:facetName rdfs:label ;
    prez:facetValue "Colobus Monkey" .
```

More comprehensive documentation are available in `docs/faceted_search.md`.

2. introduced sh:pathAlias as a means to refer to property paths (sequence, alternate etc.). These can be specified in profiles in the following form:

```turtle
sh:property [
    sh:path [
                sh:union (
                        [
                            sh:path ( prov:qualifiedAttribution prov:hadRole ) ;
                            shext:pathAlias label:role ;
                        ]
...
```
To return the path alias instead of the triples corresponding to the property path, set the `USE_PATH_ALIASES` environment variable to `true`. It is `false` by default. The path aliases can still be used for faceting without enabling this setting. If it is desired to have the path alias behaviour in some places but not others, enable the setting and create profiles with/without path aliases.

3. The parsing of SHACL Property paths now supports `sh:class`. This constrains the nodes selected by the property path to this class.

Example:
```turtle
sh:property [
    sh:path [
                sh:union (
                        [
                            sh:path ( prov:qualifiedAttribution prov:hadRole ) ;
                            sh:class ex:MyClass ;
                        ]
...
```

5. The `/profiles` endpoint now uses a DESCRIBE query to retrieve profiles. This has become necessary as the generated query to retrieve nested blank nodes to high depths (created in the sh:union constructs) is not performant in the system store oxigraph/pyoxigraph. It takes approximately 4 seconds before the change, versus sub 100 ms in Fuseki, and is now closer to 1 second after the change.

6. Query params for objects, which apparently have been missing for some time (?) are now generated in the OpenAPI docs. These are: profile, mediatype, facet_profile, subscription_key

7. fixed a bug where profiles specified by IRI required `< >` around them.


docs/faceted_search.md
```diff
@@ -0,0 +1,78 @@
...
+Faceted search is an opt in feature that is only run when the `facet_profile` query string argument is supplied. The facet profile URI, curie, or DCTERMS identifier of the profile (with or without datatype `xsd:token`) can be supplied to identify the facet profile to use.
+
+A minimal facet profile demonstrating both a direct property to facet on and one with a path alias is shown below:
+
+```turtle
+<https://prez.dev/profile/facet-by-type>
+    a prof:Profile , prez:ListingProfile ;
+    dcterms:identifier "facet-type"^^xsd:token ;
+    dcterms:title "Facet things by type and material" ;
+    dcterms:description "Allows faceting by rdf:type and sdo:material" ;
```
lalewis1, line 30: not sdo:material.

docs/path_aliases.md
```diff
@@ -0,0 +1,107 @@
...
+
+    ```turtle
+    PREFIX sh: <http://www.w3.org/ns/shacl#>
+    PREFIX dcterms: <http://purl.org/dc/terms/>
+    PREFIX prov: <http://www.w3.org/ns/prov#>
+    PREFIX shext: <http://example.com/shacl-extension#>
+
+    <http://example-profile>
+        a sh:NodeShape ;
+        sh:property [
```
lalewis1, line 30: I think these examples need to be in a sh:union

lalewis1 (approved):
just those errors in the readme, and consider adding use_path_aliases to the .env-template otherwise it looks great.
I managed to get the facets working with a path alias and without. results look good, happy to merge.

docs/faceted_search.md
recalcitrantsupplant: updated

PR #360, opened 2025-05-09
fix: hierarchy level calculation for ogc features api

recalcitrantsupplant:
url_path.replace("features/collections", "features") to url_path.replace("/features/collections", "/features") to prevent incorrectly replacing on curies that happen to end with 'features'

lets come back to this later and add some tests


ashleysommer (approved):
Looks correct to me

PR #364, opened 2025-05-19
feat: add chore to types which trigger patch releases

recalcitrantsupplant:


recalcitrantsupplant:
self approving as two line change + easiest to test by merging to main

Issue #363, opened 2025-05-19
Allow multiple uses of shext:allPredicateValues in a profile


ashleysommer:
Eg, this profile uses `shext:allPredicateValues` in 5 different places:
```turtle
    prez:BDRHumanFeatureProfile a
            prof:Profile,
            sh:NodeShape,
            prez:ListingProfile,
            prez:ObjectProfile ;
        dcterms:description "BDR Human Readable profile." ;
        dcterms:identifier "bdr-feature-human"^^xsd:token ;
        dcterms:title "BDR Feature Human Readable Profile" ;
        altr-ext:constrainsClass
            geo:Feature, rdfs:Resource , prez:CQLFilterResult , prez:SearchResult ;
        altr-ext:hasDefaultResourceFormat "application/geo+json" ;
        altr-ext:hasNodeShape
            [
                a sh:NodeShape ;
                sh:targetClass geo:Feature ;
                altr-ext:hasDefaultProfile prez:BDRHumanFeatureProfile
            ] ;
        altr-ext:hasResourceFormat
            "application/geo+json",
            "application/anot+ld+json",
            "application/ld+json",
            "application/rdf+xml",
            "text/anot+turtle",
            "text/turtle" ;
        sh:property
            [
                sh:path
                    [
                        sh:union (
                                     shext:allPredicateValues
                                     [ shext:bNodeDepth "2" ]
                                     ( geo:hasGeometry geo:asWKT )
                                     ( geo:hasDefaultGeometry geo:asWKT )
                                     ( schema:temporal shext:allPredicateValues )
                                     ( [ sh:inversePath sosa:hasFeatureOfInterest ] sosa:hasMember rdf:type )
                                     ( [ sh:inversePath sosa:hasFeatureOfInterest ] sosa:hasMember sosa:observedProperty )
                                     ( [ sh:inversePath sosa:hasFeatureOfInterest ] sosa:hasMember sosa:usedProcedure )
                                     ( [ sh:inversePath sosa:hasFeatureOfInterest ] sosa:hasMember tern:resultDateTime )
                                     ( [ sh:inversePath sosa:hasFeatureOfInterest ] sosa:hasMember sosa:phenomenonTime shext:allPredicateValues )
                                     ( [ sh:inversePath sosa:hasFeatureOfInterest ] sosa:hasMember schema:temporal shext:allPredicateValues )
                                     ( [ sh:inversePath sosa:hasFeatureOfInterest ] rdf:type )
[...]

PR #366, opened 2025-05-19
fix: create different predicates for "all predicate values"

recalcitrantsupplant:
TODO add a test

addresses #363


ashleysommer (approved):
Looks good to me.

recalcitrantsupplant:
The only other use of `?preds` I can see is under the exclusion of direct properties, this is meant to collide with all predicate values so that you can express "give me all predicates except for xyz" https://github.com/RDFLib/prez/blob/3a730683105034a4c3644ee4fb6f54abc9024e59/prez/services/query_generation/shacl.py#L541

PR #370, opened 2025-05-22
fix: redirects to existing resources without a prez:link

avillar:
Redirection to prez-ui should work even without a prez:link to the `/object?uri=...` prez-ui endpoint.

Additionally, return a 404 when the item graph is empty (ie no data found about the resource). This may not be 100% correct, though, because the item may exist but not have any data for a given profile...

### DIFF
--- a/prez/services/objects.py
+++ b/prez/services/objects.py
@@ -100,10 +100,14 @@ async def object_function(
             prez_ui_url = re.sub(r"/+$", "", settings.prez_ui_url)
             if prez_link:
                 return RedirectResponse(prez_ui_url + str(prez_link))
+            elif len(item_graph):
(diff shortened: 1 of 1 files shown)


recalcitrantsupplant:
looks fine - when/where is this getting called? when a user does a search and the class is not in the hierarchical endpoints? or when you have an IRI in the UI that doesn't have a prez link?

avillar:
Glad to hear about the prez-ui change, because that's precisely one of our use cases. 

The other is that we redirect every registered uri (eg, www.opengis.net/def/... or w3id.org/...) to /object?uri=... on the prez backend (so that Prez can either return raw RDF or redirect the user to the DE), which is where this is happening. So, even if the FE is "fixed", I think we still need this here.

PR #372, opened 2025-05-28
fix: cql nested and/or handling and alt paths within seq paths

recalcitrantsupplant:
**PR Summary:**

This PR improves SHACL and CQL query generation, focusing on better handling of complex paths and logical conditions.

**Highlights:**

* **SHACL Path Handling (a40541f, 69cee6c):**
  * Added handling of alternate paths within sequence paths, specifically for the endpoints use case in subselects. 
  * Also extended processing of alternate paths to within sequence paths in general i.e. within profiles:
    * Refactored `SequencePath` to correctly process `AlternativePath` using `UNION` blocks in `WHERE` and separate triples in `CONSTRUCT`.
  * Improved support for nested and recursive SHACL property paths.

* **CQL Logical Conditions (3431edb, af0e670):**
  * Enhanced recursive parsing of nested `AND`/`OR` conditions.
  * Fixed incorrect triple placement from `OR` branches using a depth counter.
  * Added tests to ensure proper `CONSTRUCT` and `WHERE` clause generation.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -118,14 +118,14 @@ uvloop = ["uvloop (>=0.15.2)"]
 
 [[package]]
 name = "cachetools"
-version = "5.5.2"
+version = "6.0.0"
 description = "Extensible memoizing collections and decorators"
 optional = false
-python-versions = ">=3.7"
+python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "cachetools-5.5.2-py3-none-any.whl", hash = "sha256:d26a22bcc62eb95c3beabd9f1ee5e820d3d2704fe2967cbe350e20c8ffcd3f0a"},
-    {file = "cachetools-5.5.2.tar.gz", hash = "sha256:1a661caa9175d26759571b2e19580f9d6393969e5dfca11fdb1f947a23e640d4"},
+    {file = "cachetools-6.0.0-py3-none-any.whl", hash = "sha256:82e73ba88f7b30228b5507dce1a1f878498fc669d972aef2dde4f3a3c24f103e"},
+    {file = "cachetools-6.0.0.tar.gz", hash = "sha256:f225782b84438f828328fc2ad74346522f27e5b1440f4e9fd18b20ebfd1aa2cf"},
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


ashleysommer:
Does this mean we can now do this in profiles?:

```
sh:property
    [ sh:path [
        sh:union (
           shext:allPredicateValues
           [ shext:bNodeDepth "2" ]
           ( geo:hasGeometry geo:asWKT )
           ( [ sh:inversePath sosa:hasFeatureOfInterest ] sosa:hasMember rdf:type )
           ( [ sh:inversePath sosa:hasFeatureOfInterest ] )
           ( [ sh:inversePath sosa:hasFeatureOfInterest ] [sh:alternatePath ( sosa:phenomenonTime tern:resultDateTime ) ] )
         )
      ]
    ] ;
```

Does it generate something like this with a shared obj var (like SPARQL `| ` operator)
```
CONSTRUCT
{
  ?a  sosa:phenomenonTime ?v1 .
  ?a tern:resultDateTime ?v1 .
}
WHERE 
{
  ...
  {
    ?a sosa:phenomenonTime ?v1 .
  }
  UNION
  {
   ?a tern:resultDateTime ?v1 .
  }
}
```

Or like this with separate vars?
```
CONSTRUCT
{
  ?a  sosa:phenomenonTime ?v1 .
  ?a tern:resultDateTime ?v2 .
}
WHERE 
{
  ...
  {
    ?a sosa:phenomenonTime ?v1 .
  }
  UNION
  {
   ?a tern:resultDateTime ?v2 .
  }
}
```

recalcitrantsupplant:
Currently it'll reuse the same vars for all alternates so you will end up with erroneous triples in the construct triples, this is a bug, will fix.
Also the nesting is not quite right with alternates within sequence for profiles, as per the example below.
For now with profiles please use union (mainly the change was to support endpoints which use the path syntax).

Currently:
```turtle
sh:property
    [ sh:path [
        sh:union (
           shext:allPredicateValues
           ( sosa:hasFeatureOfInterest [ sh:alternativePath ( sosa:phenomenonTime tern:resultDateTime ) ] )
         )
      ]
    ] ;
```
generates:

```sparql
    {
        {
            ?prof_1_node_2 <https://www.w3.org/TR/vocab-ssn/phenomenonTime> ?prof_1_node_3
        }
        UNION
        {
            ?prof_1_node_2 <http://w3id.org/tern/ontologies/tern/resultDateTime> ?prof_1_node_3
        }
    }

    {
        ?focus_node ?preds ?vals
    }
    UNION
    {
        ?focus_node <https://www.w3.org/TR/vocab-ssn/hasFeatureOfInterest> ?prof_1_node_2
    }
```
The hasFeatureOfInterest triple should be adjacent to the time union triples not separate.

recalcitrantsupplant:
> This could be implemented by simply aliasing to the first predicate, in the CONSTRUCT query:

Yes - not sure if you were referring to this but you can use the shext:pathAlias for this purpose:

https://github.com/RDFLib/prez/blob/b34e847193b5d918e37a31444c2a8ab6ed74a70c/tests/test_shacl_sequence_alternative_paths.py#L92

From subsequent discussions in the SHACL WG I believe sh:values should/could be used instead of shext:pathAlias.

lalewis1 (approved):
Have tested with various nested and/or CQL queries and the generated SPARQL is good, couldn't break it.

have also tested with various alternativePaths by overriding the default profile profile and again, couldn't break it.

LGTM

recalcitrantsupplant:
> Have tested with various nested and/or CQL queries and the generated SPARQL is good, couldn't break it.

Cheers, NB the generated SPARQL should always be valid (unless there's a bug with sparql-grammar-pydantic) just potentially semantically incorrect.

Noting the known issues and merging https://github.com/RDFLib/prez/issues/374

PR #375, opened 2025-06-12
feat: add shacl filterShape support for conditional triple pattern emission

recalcitrantsupplant:
Implements support for SHACL sh:filterShape to enable conditional triple pattern emission based on sibling property constraints. This allows profiles to emit triples only when the subject node satisfies specific filter conditions using standard SHACL vocabulary.
The implementation adds _parse_filter_shape() method to extract filter conditions and enhances path processing to handle filter shapes. Filter conditions are automatically flattened and applied directly to the focus node, with context guards preventing infinite recursion during processing.
Example Usage:
```turtle
sh:property [
  sh:path ( sosa:hasResult rdf:value ) ;
  sh:filterShape [
    sh:property [
      sh:path ( [ sh:inversePath rdf:value ] [ sh:inversePath sosa:hasResult ] ) ;
      sh:property [
        sh:path sosa:observedProperty ;
        sh:hasValue ex:Temperature
      ]
    ]
  ]
] .
```
Generated SPARQL:
```sparql
?focus_node sosa:observedProperty ex:Temperature .
?focus_node sosa:hasResult ?prof_1_node_1 .
?prof_1_node_1 rdf:value ?prof_1_node_2 .
```
Currently supports sh:hasValue constraints with sh:in support planned.

### DIFF
--- a/prez/services/query_generation/shacl.py
+++ b/prez/services/query_generation/shacl.py
@@ -428,10 +428,6 @@ class PropertyShape(Shape):
 
             bnode_class = next(self.graph.objects(subject=pp, predicate=SH["class"]), None)
 
-            bnode_filter_shape_uri = next(self.graph.objects(subject=pp, predicate=SH.filterShape), None)
-            if bnode_filter_shape_uri:
-                final_filter_shape_tssp_list.extend(self._parse_filter_shape(bnode_filter_shape_uri))
-
             nested_path_node = next(self.graph.objects(subject=pp, predicate=SH.path), None)
             if nested_path_node:
                 path_to_parse = nested_path_node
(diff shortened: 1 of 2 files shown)


lalewis1:
thanks David, so much prez dev work lately.

I noticed the generated count query does not take the filtered results into account. 
and the sh:hasValue does not work with literal values (not sure if you care about this)

as in,
```shacl
...
sh:path ex:prop ;
sh:hasValue "propertyValue" ;
...
```

comes out in the sparql as
```sparql
?focus_node <http://example.org/prop> <propertyValue>
```

otherwise, looks good.

recalcitrantsupplant:
Cheers I'll add fixes for those issues.

Specifically for the count, it's meant to be a count of focus nodes, so I think the solution is to always (hardcode) include at least `rdf:type` in the properties, so there is always at least one triple for a focus node returned. At present if people write profiles which don't "match" the data, they get a count of focus nodes which confusingly doesn't match the data that is returned.
There's a similar problem when creating GeoJSON for focus nodes that do not have a geometry.
I'll look into using POINT EMPTY or MULTIPOLYGON EMPTY (which are valid WKT) to see if it's a viable workaround.

PR #376, opened 2025-06-16
fix: update sparql grammar to terminate all triples same subject path with fullstop

recalcitrantsupplant:
Update sparql grammar pydantic. TriplesBlock will now always produce a full stop "." per triple (including cases where it isn't required). This is to cover cases where multiple TriplesBlocks with a single triple are used together and previously did not terminate their single triples with a "."

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -130,14 +130,14 @@ files = [
 
 [[package]]
 name = "certifi"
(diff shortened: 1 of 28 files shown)


lalewis1 (approved):
Looks good thanks. approved.

Issue #351, opened 2025-04-11
SPARQL/Data Out - JSON-LD Framing


edmondchuc:
In addition to the current SPARQL Endpoint's current behaviour, have the option for the user to supply a JSON-LD Frame to frame the CONSTRUCT query results on the way out as an additional query parameter.

Example: `/sparql?query=construct....&frame=....`

recalcitrantsupplant:
We could enable this globally.
How would it work with profiles/mediatypes - just override any requested/default mediatypes with json-ld?

edmondchuc:
I thought profiles only worked on resources, or do they apply to the SPARQL endpoint as well?

recalcitrantsupplant:
Profiles do not apply to the SPARQL endpoint. They do apply to other endpoints though, which I'm suggesting this feature could cover as well.

For the SPARQL endpoint, we would need to detect the query type which we are not currently doing.

edmondchuc:
Ok, I understand now. 

The way I see it is, the profiles pull in data to the response graph closure and framing shapes the data.

In the case of resource items, I am not sure if the media type is necessary or not. The response will always return `application/ld+json` when a `frame` query parameter is supplied. The endpoint will process things as it does now (applying profiles, pulling data into the graph closure, etc) and then as a last step, convert it to JSON-LD and apply framing to it.

What do you think?

recalcitrantsupplant:
Yes I can see it working that way on the non SPARQL endpoints; yes the mediatype isn't necessary if a frame is supplied, but connegp would determine one anyway which would then need to be ignored (which seems like a logical tradeoff).

PR #383, opened 2025-07-03
refactor: remove dropdown jsonld and csv specific code

### LINKED ISSUE #382: Bug in renderer, getting annotations for when profile is "https://w3id.org/profile/dd"
When requested profile is "https://w3id.org/profile/dd", the renderer codepath calls `return_annotated_rdf()` with incorrect arguments.
https://github.com/RDFLib/prez/blob/14991e943581155d2c18f59e0851e9d7fc5f9e37/prez/renderers/renderer.py#L62

Arguments should be `(Graph, Repo, Repo)` like this one further down in the renderer codepath:
https://github.com/RDFLib/prez/blob/14991e943581155d2c18f59e0851e9d7fc5f9e37/prez/renderers/renderer.py#L94

ashleysommer:
Fixes #382


recalcitrantsupplant:
@edmondchuc do we still need the API to return CSV/special JSON LD, or can we leave it to FE / other clients to create the CSV? the standard listing endpoint profiles can do "skos:Concept instances are listed, alphabetically by label, that their labels (skos:prefLabel) are present and that if they have a broader skos:Concept, that too is given" in a paginated alphabetically ordered manner, the output would be RDF. The code here at present won't do what it was intended for as it was written for version 3 profiles https://github.com/RDFLib/prez/blob/45ad323bc8c7d6fe3b9c0c7a18adddf99c6a68c1/prez/renderers/json_renderer.py#L66

edmondchuc:
Hi @recalcitrantsupplant, I think we can probably move existing clients that use the DD profile to use the SPARQL endpoint instead. I quite like the pattern of providing other systems a URL with the SPARQL query as a query parameter using GET.

Another approach I've always wanted to explore is having an endpoint that accepts both a SPARQL construct query and a JSON-LD frame. That would let us shape the response structure to suit each client’s needs by adjusting just the query and the frame/context.

Both options seem more flexible and powerful compared to the current DD profile.

edmondchuc:
Mentioning this here for future discussions: https://github.com/RDFLib/prez/issues/351

Issue #381, opened 2025-06-30
[Feature] Use Oxigraph to parse SPARQL results, and to serialize RDF responses


ashleysommer:
I know we've discussed this before, this is one of the last big bottle necks for Prez, especially when it comes to largish responses.

After running some tests today, I have some results which show the impact of RDFlib's parsing and serialization on API response times.
All requests are done via the /cql endpoint with the following filter:
```json
{
    "op": "and",
    "args": [
        {
            "op": "in",
            "args": [
                {
                    "property": "nsl-name-usage"
                },
                [
                    "https://id.biodiversity.org.au/instance/apni/7646321"
                ]
            ]
        },
        {
            "op": "=",
            "args": [
                {
                    "property": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
                },
                "http://www.opengis.net/ont/geosparql#Feature"
                
            ]
        }
    ]
}
```

### Scenario 1, with profile="minimal-feature-listing"
Medium response size, limit=100 features
1) Raw SPARQL response times: 0.6s (1,216 triples)
2) API response format text/turtle: 0.7s (9.9kb)

### Scenario 2, with profile="minimal-feature-listing"
Large response size, limit=500 features
1) Raw SPARQL response times: 0.6s (6,038 triples)
2) API response format text/turtle: 3.12s (46.4kb)

### Scenario 3, with profile="minimal-feature-listing"
V-Large response size, limit=2000 features
1) Raw SPARQL response times: 0.7s (23,848 triples)
2) API response format text/turtle: 11.21s (182kb)

### Scenario 4, with profile="feature-attributes-human"
Medium response size, limit=100 features
1) Raw SPARQL response times: 0.6s  (4,941 triples)
2) API response format text/turtle: 2.64s (308kb)

### Scenario 5, with profile="feature-attributes-human"
Large response size, limit=500 features
1) Raw SPARQL response times: 0.6s  (24,713 triples)
2) API response format text/turtle: 11.9s (1,538kb)

### Scenario 6, with profile="feature-attributes-human"
Large response size, limit=2000 features
1) Raw SPARQL response times: 0.7s  (97,810 triples)
2) API response format text/turtle: 44.8s (6,130kb)

_Actual SPARQL query response times are gathered by using the "\_mediatype=application/sparql-query" feature to get the generated SPARQL query, then running that query in the GraphDB workbench._

Note, I did not use an anot+ media type, so to not complicate the results with the extra parallel query.
[...]

PR #384, opened 2025-07-03
feat: oxigraph sparql responses

ashleysommer:
This is a first-draft attempt at an implementation for #381 

Currently implemented is:

* Add Oxigraph output support for backend Repo types (Oxrdflib, Pyoxigraph, and RemoteSparql Repos)
  - This means all triple-returning queries (CONSTRUCT or DESCRIBE queries) can return an Oxigraph store instead of an RDFlib graph, regardless of backend type
  - For remote-sparql kind, this also uses the fast Pyoxigraph parser to parse the sparql-response RDF.
* Add Oxigraph serialization support to the render pipeline
  - Includes link-generation handling where required
    - Extracts unique objects from the Oxigraph store, finds their classes
    - runs lookups for the class links, adds the links back to the oxigraph store
  - Additing annotations if required
    - Finds all unique subjects, objects, predicates, and datatypes in the Oxigraph store
    - Runs lookups to get labels/annotations for these unique uris, adds the annotations back to the oxigraph store.
* Adding prefixes from the prefix-cache graph to the Oxigraph serializer output
* Also adds support for the json-dropdown output profile
  - I don't know if this is still used, or what it even does, but I was able to make it compatible with Oxigraph output-graph case.
* Enabled the above features in the listing_function endpoint, and got tests passing

Not ported yet:
* ~Ogc-features-listing-function endpoint~  done
* ~Object endpoint~ done

This is a draft for feedback, to see if this is on the right path, and to confirm this is the correct approach.


ashleysommer:
Note, some tests are now failing due to the Oxigraph serializer thinking one or more of the provided prefixes from the prefix-cache graph are invalid.

```python
FAILED tests/test_endpoints_cache.py::test_purge_cache - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
FAILED tests/test_endpoints_cache.py::test_cache - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
FAILED tests/test_endpoints_catprez.py::test_catalog_listing_anot - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
FAILED tests/test_endpoints_concept_hierarchy.py::test_concept_hierarchy_top_concepts - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
FAILED tests/test_endpoints_concept_hierarchy.py::test_concept_hierarchy_narrowers - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
FAILED tests/test_endpoints_ok.py::test_visit_all_links - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
ERROR tests/test_endpoints_catprez.py::test_catalog_anot - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
ERROR tests/test_endpoints_catprez.py::test_catalog_no_mediatype - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
ERROR tests/test_endpoints_catprez.py::test_lower_level_listing_anot - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
ERROR tests/test_endpoints_spaceprez.py::test_dataset_anot - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
ERROR tests/test_endpoints_spaceprez.py::test_feature_collection - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
ERROR tests/test_endpoints_spaceprez.py::test_feature - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
ERROR tests/test_endpoints_spaceprez.py::test_feature_listing_anot - ValueError: Invalid prefix d IRI '://d/', No scheme found in an absolute IRI
```

I don't know whats causing that. Seems like Oxigraph thinks there is a prefix named `"d"` with a namespace of `"://d/"` and it complains the namespace has no schema component. 
I don't know how this prefix is getting in there, its not in the prefix-cache and if I print out the contents of the prefix cache immediately before calling `store.dump()` that prefix is not in there.

ashleysommer:
Also, needing to be completed, is adding Oxigraph support to the rdf2geojson converter.

ashleysommer:
Update, rdf2geojson v0.8.0 now works with Oxigraph stores. The only difference in the API surface is you also need to pass in a namespace_manager to the `convert` function if the graph is a Oxigraph store (because oxigraph stores do not have their own namespace manager).
I've got some more fixes in now, the test failure above is no longer blocking.

prez/dependencies.py
```diff
@@ -162,9 +173,10 @@ async def load_annotations_data_to_oxigraph(store: Store):
     g = Dataset(default_union=True)
     for file in (Path(__file__).parent / "reference_data/annotations").glob("*"):
         g.parse(file)
+    # Whats the point of loading this into an RDFLib dataset them serializing it again
+    # to then load it into the oxigraph store?
```
recalcitrantsupplant, line 177: No point - probably a copy paste error, can be simplified

prez/services/annotations.py
```diff
@@ -41,6 +44,44 @@ async def get_annotations(terms_and_dtypes: Set[URIRef], repo: Repo, system_repo
...
+        system_repo (Repo): An instance of the Repo class with the Prez system graph.
+
+    Returns:
+        annotations_store (OxiStore): A oxigraph store containing the processed terms and their data types.
+    """
+    annotations_store = OxiStore()
+    oxi_default_graph = OxiDefaultGraph()
+    cache = caches.get("default")  # This always returns the SAME instance
+    # Cache always uses URIRefs, so we convert OxiNamedNode to URIRef
+    results = await cache.multi_get(list(URIRef(t.value) for t in terms_and_dtypes))
```
ashleysommer, line 65: The aiocache uses the pickle serializer, to store the mapping of `URIRef` to `frozenset((p,o))` where p and o are RDFLib terms. We cannot have the equivelent separate cache for mapping Oxigraph `NamedNode` to `frozenset((p,o))` where p and o are Oxigraph terms. Because Oxigraph terms are not picklable.

So to use the annotations cache, we need to convert the Oxigraph terms to RDFlib terms. And when loading from the cache, we convert them back again (see below).

One upside to doing it this way is it means `get_annotations_for_oxigraph` and `get_annotations` functions called from different codepaths can both use and benefit from the same annotations cache.

prez/dependencies.py
ashleysommer: This is resolved in https://github.com/RDFLib/prez/pull/384/commits/455bb746fd982bb446593eb001857fcbef34e567

ashleysommer:
Thanks @recalcitrantsupplant 
See also related https://github.com/oxigraph/oxigraph/issues/1344 and https://github.com/oxigraph/oxigraph/issues/1345

ashleysommer:
@recalcitrantsupplant  In your recent fix for CollectionJSON, pyoxigraph `NamedNode` was already imported as `OxiNamedNode`, you've added a second import in your commit.

recalcitrantsupplant:
When requesting the alternates profile, there there are two queries generated, a "data" query and a count query.
If I set use_oxigraph = True here:
https://github.com/RDFLib/prez/blob/16f0c7fc0411c8b92410fe9da665cc5fb2807525/prez/services/listings.py#L141
Then the results are gathered here:
https://github.com/RDFLib/prez/blob/39ffcec1b92385eaec87ab4f5f91253b633bb5d4/prez/repositories/base.py#L58

This hangs for me, it never returns. With use_oxigraph = False, it returns fine. I'm not sure we should be calling gather with into_graph being the same instance for both. For some reason I don't get this issue when not using the alternates profile, i.e. a regular listing at /catalogs despite there being >1 query as well.

ashleysommer:
@recalcitrantsupplant 
> This hangs for me, it never returns. 

I think it might be due to the threading used in the Repos implementations. Its probably a bad idea to pass a single pyoxigraph Store into more than one thread at once, as its not a Python object.
I'm not sure if the threading is really required in those repos, but that might need to be revisited.

recalcitrantsupplant:
Sounds good, thanks for the analysis in #401, I've known this area of the code is in need of a refactor. I think the idea of collecting the queries by nodeshape is good - I'd even previously considered that it could go into a single SPARQL query with union - even a query per nodeshape as I think you're doing would obviously be an improvement too. The original implementation was not able to generate the SPARQL path expressions (*/^ etc.), so it made joining different queries hard. I don't think the same limitations are there now though.

> 10 minutes to generate links using the cache.

10 seconds?

ashleysommer:
> > 10 minutes to generate links using the cache.
>
> 10 seconds?

No, 10 **minutes** is correct. This is discussed in #401, and is the main motivator for this change. When using an RDFLib graph as the link cache, the cache lookup uses `graph.quads((None, None, None, uri))`.

GraphDB doesn't have a context index, so it goes through every triple in the graph and checks if its context matches that given URI. It then needs to do that 10,000 times for each URI in the current response.

When there are a large number of links stored (eg, 10,000 from a previous lookup), there will be around 30,000 triples in the cache (for items and members links), and with 10,000 links in the current response, that means its doing 300,000,000 checks on the link cache for a single response. Thats why it takes so long.

Switching the links cache to an oxigraph store means the lookup is now `store.quads_for_pattern(None, None, None, uri_node)`, and oxigraph _does_ have a context index, that means it can give only the saved triples that are relevant to that URI. Thats why in my testing the cached lookups for 10,000 items goes from 10 minutes, to 19 seconds.

recalcitrantsupplant:
OK I've re-read more carefully, got it.

recalcitrantsupplant:
I think I've faithfully replicated the store vs quads approaches in the benchmark, do you want to have a look and run against any backends you will use.

This is against one of the BDR repos:

| Test Run             | 10         | 100        | 1000       | 10000      | 30000      |
|----------------------|-----------|-----------|-----------|-----------|-----------|
| Run 1 - Quads        | 0.002      | 0.001      | 0.006      | 0.064      | 0.060      |
| Run 1 - Stores       | 0.003      | 0.001      | 0.004      | 0.046      | 0.044      |
| Run 2 - Quads        | 0.001      | 0.001      | 0.005      | 0.057      | 0.058      |
| Run 2 - Stores       | 0.003      | 0.001      | 0.005      | 0.042      | 0.049      |
| Run 3 - Quads        | 0.001      | 0.001      | 0.006      | 0.058      | 0.051      |
| Run 3 - Stores       | 0.003      | 0.001      | 0.005      | 0.045      | 0.048      |
|----------------------|-----------|-----------|-----------|-----------|-----------|
| Average - Quads      | 0.001      | 0.001      | 0.006      | 0.060      | 0.057      |
| Average - Stores     | 0.003      | 0.001      | 0.004      | 0.044      | 0.047      |

recalcitrantsupplant:
I think most responses would be close to >100 triples, and the difference would becomes more noticeable at larger numbers, so for these reasons I think switching to your store with lock implementation is the way to go. Do you agree @ashleysommer - if so we can revert my last commit and probably cherry pick from your branch?

Otherwise I think it's good to merge. I think there's potentially a few other places RDFLib usage could now be replaced with Oxigraph, but this could be in follow up PRs.

recalcitrantsupplant:
There was a lock deletion bug I encountered above which I put a fix in for here: https://github.com/RDFLib/prez/pull/384/commits/364b9431629bf72ad85443363cbfbaa9a8e7625d

[error_1.txt](https://github.com/user-attachments/files/21346074/error_1.txt)
[error_2.txt](https://github.com/user-attachments/files/21346087/error_2.txt)

> there might be still some edge cases where Repos are trying to write into the same Oxigraph store at the same time

I'll give this some more thought, I'm not sure exactly where/when it would happen, perhaps around the annotation store?

ashleysommer:
I updated the Branch at https://github.com/RDFLib/prez/pull/404 with the fix for the logic bug, and its working correctly now

recalcitrantsupplant:
The tests pass locally for me; the CI tests now have the error I was previously getting locally.

Thoughts on a global lock manager, something like:

```python
from weakref import WeakKeyDictionary
from threading import Lock
from pyoxigraph import Store
from fastapi.concurrency import run_in_threadpool

class LockManager:
    def __init__(self):
        self._locks = WeakKeyDictionary()
        self._creation_lock = Lock()

    def get_lock(self, obj):
        if obj not in self._locks:
            with self._creation_lock:
                if obj not in self._locks:
                    self._locks[obj] = Lock()
        return self._locks[obj]

repo_locks = LockManager()

class PyoxigraphRepo(Repo):
    def __init__(self, pyoxi_store: Store):
        self.pyoxi_store = pyoxi_store

    def _sync_rdf_query_to_oxigraph_store(self, query: str, into_store: Store | None) -> Store:
        results = self.pyoxi_store.query(query)
        
        if into_store is not None:
            with repo_locks.get_lock(into_store):
                return self._handle_query_triples_results(results, into_=into_store)
        else:
            s = Store()
            return self._handle_query_triples_results(results, into_=s)

    async def rdf_query_to_oxigraph_store(self, query: str, into_store: Store | None = None) -> Store:
        return await run_in_threadpool(self._sync_rdf_query_to_oxigraph_store, query, into_store)
```

ashleysommer:
@recalcitrantsupplant 
Sorry, I forgot to actually push the change to the branch in #404, so what you merged was not the fix (and it removed your fix too).
I'll push the missing change to this branch.

ashleysommer:
Yes, I'd still like to see a global "into_store" lock, that works on all repos. Maybe not as part of this PR, done as a separate change.
I have an idea for how to implement it in a relatively simple way (or your proposed method looks good too).

recalcitrantsupplant (approved):

Issue #385, opened 2025-07-03
SKOS Collection member pagination support


edmondchuc:
Provide `skos:member` pagination support in the same way we do for `skos:hasTopConcept` and `skos:narrower`.

recalcitrantsupplant:
To clarify is this specifically for skos:Collections which have skos:member another skos:Collection?

recalcitrantsupplant:
This can be done with the existing ability to define custom endpoints, and the default endpoints already provide this here (under a top level dcat:Catalog):
https://github.com/RDFLib/prez/blob/142acd1e2131e582b8d831eed5894d8141a2c8df/prez/reference_data/endpoints/data_endpoints_default/default_endpoints.ttl#L58
https://github.com/RDFLib/prez/blob/142acd1e2131e582b8d831eed5894d8141a2c8df/prez/reference_data/endpoints/data_endpoints_default/default_endpoints.ttl#L73

recalcitrantsupplant:
I'll close this issue, we can create issues in the prez demo repo for specific configurations if they're not covered in the main repo here. https://github.com/Kurrawong/prezdemo/tree/main/examples

PR #386, opened 2025-07-03
feat: add prez:OGCSKOSCollectionObjectProfile and update prez:OGCSchemesObjectProfile

edmondchuc:
### SKOS Collection Object Profile

This new profile `prez:OGCSKOSCollectionObjectProfile` constrains on the class `skos:Collection` and allows prez to include `skos:member` values in its responses.

Note that the `skos:member` values do not support pagination like it does with concept schemes. This will have performance issues with very large collections. This performance issue is tracked here https://github.com/RDFLib/prez/issues/385. We can follow up on this issue when there's a requirement, otherwise this PR should keep the ICSM WG happy!

Screenshot 2025-07-03 at 12 54 53 pm

### Concept Scheme Profile Update
This PR also updates `prez:OGCSchemesObjectProfile` to return the `skos:inScheme` inverse relationships from `skos:Collection` objects pointing to the focus node (`skos:ConceptScheme`). This is required to allow prez ui to list the relevant collections within the "vocab".

### Outstanding Tasks

- [ ] - fix `sh:class` to SPARQL query filter - once this bug is fixed, the tests in this PR should pass


recalcitrantsupplant (approved):
Approved noting we'll fix the sh:class restriction in profiles so that not all skos:inScheme related resources are included, as including all concepts can return a high number of results.

PR #388, opened 2025-07-04
feat: add on-disk data store for pyoxigraph

edmondchuc:
This PR adds support for using an on-disk store with the pyoxigraph data store backend.

If the environment variable `PYOXIGRAPH_DATA_DIR` is set, the data store will treat the specified location as a valid oxigraph data directory and operate in persistent (on-disk) mode instead of the default in-memory mode.

### Logging behaviour change

Not sure why log propagation was originally disabled. Without log propagation, I couldn't use caplog to capture the logs in tests.

Let me know if this needs to be added back in. We can make it configurable via settings so that we can enable it in tests and disable it in deployments.

### DIFF
--- a/prez/app.py
+++ b/prez/app.py
@@ -102,10 +102,13 @@ async def lifespan(app: FastAPI):
     log = logging.getLogger("prez")
     log.info("Starting up")
 
-    if app.state.settings.sparql_repo_type == "pyoxigraph":
+    if app.state.settings.sparql_repo_type == "pyoxigraph_memory":
         app.state.pyoxi_store = get_pyoxi_store()
         app.state.repo = PyoxigraphRepo(app.state.pyoxi_store)
         await load_local_data_to_oxigraph(app.state.pyoxi_store)
+    elif app.state.settings.sparql_repo_type == "pyoxigraph_persistent":
+        app.state.pyoxi_store = get_pyoxi_store()
+        app.state.repo = PyoxigraphRepo(app.state.pyoxi_store)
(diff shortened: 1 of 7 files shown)


recalcitrantsupplant:
How about:
```
sparql_repo_type: Literal["remote", "pyoxigraph_store", "pyoxigraph_from_rdf"] = "remote"
pyoxigraph_data_dir: str = None
```
And we deprecate:
~local_rdf_dir: str = "rdf"~
`pyoxigraph_data_dir` would be used for either a pyoxigraph store location (`sparql_repo_type = pyoxigraph_store`), or a directory with rdf files (`sparql_repo_type = pyoxigraph_from_rdf`).
This would mean we can keep two config options instead of a new one, and that the repo type is explicit through `sparql_repo_type`.
Technically this is breaking, I do not believe anyone is using the pyoxigraph option in production, I use it locally sometimes.

edmondchuc:
@recalcitrantsupplant @lalewis1, I've just added a commit to address David's comment.

Note: I've made one functional change to the glob pattern used to load the RDF files for the pyoxigraph memory store. It now loads all `*.ttl` in the target directory and all sub-directories.

Also I've opted for `pyoxigraph_persistent` and `pyoxigraph_memory` as the repo types instead of `pyoxigraph_store` and `pyoxigraph_from_rdf`. The distinction seems clearer this way where one uses disk storage, while the other is in-memory, even if it loads data from RDF files.

recalcitrantsupplant:
> I've opted for pyoxigraph_persistent and pyoxigraph_memory
Sounds good to me

PR #390, opened 2025-07-09
fix: pyoxigraph persistent store lock contention

edmondchuc:
Each request was creating a new pyoxigraph `Store` and trying to access the same directory. This PR ensures only one store is created and reused across requests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+pyoxigraph_data_dir/
+
 venv/
 __pycache__/
 .venv/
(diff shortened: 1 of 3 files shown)


recalcitrantsupplant:
Can the same pattern that is used for the other pyoxigraph store be used where it is populated on startup in the lifespan event (in app.py)?

edmondchuc:
Thanks, I've just pushed a commit to address that issue.

recalcitrantsupplant (approved):

PR #392, opened 2025-07-09
fix: don't recreate remotesparql repo and http_async_client on every request

### LINKED ISSUE #391: get_async_http_client runs for every request, and takes an unuaually long time
While trying to track down some odd performance issues during the development of feature #384 
I added some instrumentation using PyInstrument (see the guide [here ](https://stackademic.com/blog/profiling-fastapi-for-ultimate-performance-tuning))

This is what the flamegraph looks like for a single request, with dependencies and annotations _uncached_, this is the first request after starting the application. (click for larger image)

Image

This is a 4 second timeframe, notice the first 1.05s second (a quarter of the whole request) is taken up by "get_async_http_client" in the fast-api dependencies.

Now this is the flamegraph for a subsequent request, exactly the same url and parameters, this time dependencies and annotations are _cached_. (click for larger image)

Image

This period is now 2.9s, but "get_async_http_client" is there again, it takes up 1.05s again, this time its more than a third of the response time.

I've checked the code and it really seems like Fast-API should be treating this as a cached dependency, so I don't know why its recreating the client for every request.  Also, I don't know why creating a httpx client should take 1.05s, normally its a very fast operation, it sould be 1ms max.

ashleysommer:
Fixes #391 
This prevents the creation of a new `http_async_client` for every request.

1) Ensure some of the cached state objects that get set on the main FastAPI app also gets set on the mounted Sub-Apps.
2) Remove "get_async_http_client" dependency from the "get_data_repo" dependency. This stops a new client from being created every request
3) In `get_data_repo()` function, try to use the saved `app.state.repo`, and only fall back to creating the repo if that saved state object is not available

### DIFF
--- a/prez/app.py
+++ b/prez/app.py
@@ -112,12 +112,17 @@ async def lifespan(app: FastAPI):
             mount_app = r.app
             if isinstance(mount_app, (Starlette, FastAPI)) and mount_app is not app:
                 mounted_apps.append(mount_app)
-    if app.state.settings.sparql_repo_type == "pyoxigraph":
+    if app.state.settings.sparql_repo_type == "pyoxigraph_memory":
         app.state.pyoxi_store = pyoxi_store = get_pyoxi_store()
         for mounted_app in mounted_apps:
             mounted_app.state.pyoxi_store = pyoxi_store
         app.state.repo = repo = PyoxigraphRepo(pyoxi_store)
         await load_local_data_to_oxigraph(pyoxi_store)
+    elif app.state.settings.sparql_repo_type == "pyoxigraph_persistent":
+        app.state.pyoxi_store = pyoxi_store = get_pyoxi_store()
+        for mounted_app in mounted_apps:
+            mounted_app.state.pyoxi_store = pyoxi_store
+        app.state.repo = repo = PyoxigraphRepo(pyoxi_store)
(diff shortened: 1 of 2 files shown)


edmondchuc:
Just a quick question: if we had module level globals instead of app.state level globals, would that help with the sub-app mounting issue?

ashleysommer:
Yeah, perhaps it would work around this but I'm not a fan of module level globals.  I think this solution to sync the base states at startup in the lifespan is an okay solution, but it does mean the app states could get out of sync if anything is added/changed to one app during runtime.

ashleysommer:
I worked out why creating the http client each request was causing such a slowdown in my testing.
Its actually 2 reasons.
1) Windows. My work laptop is Windows 11, and for some reason creating a httpx Client or AsyncClient takes longer, something to do with accessing the OS certificate stores.
2) Proxy. My work laptop has an always-on SIG (secure internet gateway) proxy. httpx library detects that and needs to read the proxy information from the OS to set up a different transport layer for requests which go through the proxy. This is slow on Windows 11, it can take 500ms-1000ms for the OS to respond when asked for the proxy information.

recalcitrantsupplant:
Great - good to know the cause and that it shouldn't be an issue with cloud deployments.

PR #397, opened 2025-07-14
fix: try to use request accept mediatype in fallback profile

ashleysommer:
When a specific profile isn't requested, and prez cannot find a profile to use for the object class, it falls back to hardcoded profiles  - open-object for Object endpoint, mem (Members) for Listing endpoint.

The problem is this fallback profile uses a single hardcorded mediatype "text/anot+turtle". Even if you have header "Accept: text/turtle" in the request or "?_mediatype=text/turtle" in the QSA, prez will still always use "text/anot+turtle".

https://github.com/RDFLib/prez/blob/cc8c3cac4ab3bc0580e65539f4568c43be28512e/prez/services/connegp_service.py#L217-L224

https://github.com/RDFLib/prez/blob/cc8c3cac4ab3bc0580e65539f4568c43be28512e/prez/services/connegp_service.py#L226-L233

In this PR:
Instead of using hardcoded mediatypes in fallback profiles, try to use the one that is requested if it matches one of the mediatypes allowed on the fallback profile.

### DIFF
--- a/prez/services/connegp_service.py
+++ b/prez/services/connegp_service.py
@@ -1,6 +1,5 @@
 import logging
 import re
-import time
 from enum import Enum
 from textwrap import dedent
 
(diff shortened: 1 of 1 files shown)


prez/services/connegp_service.py
```diff
@@ -212,22 +211,51 @@ async def _get_available(self) -> list[dict]:
...
+            else:
+                fallback_mediatype = None
             if self.listing:
+                # known available mediatypes for the 'mem' profile
+                _available_mem_types = [
+                    "application/ld+json" ,
+                    "application/anot+ld+json" ,
+                    "application/rdf+xml" ,
+                    "text/anot+turtle" ,
+                    "text/turtle"
```
ashleysommer, line 227: The list of available types to match on are pulled from the actual profile definition. https://github.com/RDFLib/prez/blob/cc8c3cac4ab3bc0580e65539f4568c43be28512e/prez/reference_data/profiles/prez_default_profiles.ttl#L114-L119

recalcitrantsupplant:
Sounds good - plan to merge #398 & #384 first

recalcitrantsupplant:
future work captured in https://github.com/RDFLib/prez/issues/409

Issue #396, opened 2025-07-14
[Question]


ashleysommer:
Why does define "sdoprof" prefix for https://schema.org/ ?

https://github.com/RDFLib/prez/blob/cc8c3cac4ab3bc0580e65539f4568c43be28512e/prez/reference_data/prefixes/standard.ttl#L18-L21

I know RDFLib previously used "sdo" but recent discussions have been to switch to using "schema" prefix.

I've been using "schema" prefix in all my RDFLib applications for the last 18 months.

Issue #395, opened 2025-07-14
[Feature Request] Add option to not use baked-in all.file.vann.ttl prefixes


ashleysommer:
During startup, prez pre-loads a list of 2969 pefix-namespace pairs from the all.file.vann.tll that is part of prez.
It then loads the local standard.tll prefixes file (with 13 prefix-namespace pairs), and any defined prefix-namespaces from the remote store.

While experimenting with replacing RDFlib graph output serialization with PyOxigraph, one of the main differences in the serializer is that the oxigraph Turtle serializer will output ALL provided prefixes in the namespace manager, regardless of whether those prefixes are used in the output graph.

This often results in API responses of Turtle data with 2900 prefixes, and less than 100 triples of real data. This uses excess bandwidth when sending responses over the network, and extra filespace when storing the output files.

I wonder if there could be an setting `INCLUDE_VANN_PREFIXES` I could toggle off to prevent the inclusion of "all.file.vann.ttl" file in the prefixes cache.

ashleysommer:
Another thing I should point out, I noticed in "standard.ttl" the values for `vann:preferredNamespaceUri` are URIs eg `<https://schema.org/>` where in "all.file.vann.tll" the values for `vann:preferredNamespaceUri` are all literal strings, eg "https://schema.org/" . I suppose both work?

PR #398, opened 2025-07-14
feat: prefix behaviour changes

recalcitrantsupplant:
This PR is intended to improve/clean up prefix handling prior to merging https://github.com/RDFLib/prez/pull/384

- Remove vann namespace prefixes from the repo altogether.
- Remove rdflib default prefixes.
- Only those prefixes in `prez/reference_data/prefixes/*.ttl` are loaded on startup now, along with any remote prefix declarations. This simplifies prefix handling, improves response times slightly, and accommodates the current behaviour of pyoxigraph where all prefixes passed to the serialise method were included in a response, regardless of whether they were used. This bloated responses as ~3000 prefixes were included.

These changes mean users will potentially see auto generated prefixes (if previously using prefixes covered by vann or rdflib default) unless they specify prefixes through the existing unchanged prefix declaration format using vann.
To find missing namespace prefixes please see:
https://github.com/RDFLib/prez/blob/main/docs/development.md#generating-prefixes
A future feature is planned to support specifying these more easily in a "configuration mode"

Resolves https://github.com/RDFLib/prez/issues/396
Resolves https://github.com/RDFLib/prez/issues/395

Other small unrelated changes:
- update deps
- remove xfail on OGC Features test that passed but had a test suite issue


ashleysommer:
> These changes mean users will potentially see auto generated prefixes (if previously using prefixes covered by vann or rdflib default) unless they specify prefixes through the existing unchanged prefix declaration format using vann.

The oxigraph serializer does not auto-generated prefixes (eg, like RDFlib does "ns1", "ns2", etc).
It will simply leave the URIs in their full form.

I do have a question about curies. When Prez generates a curie for a known class type (so it can genreate Links and so you can refer to that identifier in URLs), does it add those curie prefixes into the prefix namespace cache? Or are they kept separate? Or is this what you were referring to when you said there will be more auto-generated prefixes (generated by prez curie maker).

recalcitrantsupplant:
> I do have a question about curies. When Prez generates a curie for a known class type (so it can populate Links and so you can refer to that identifier in URLs), does it add those curie prefixes into the prefix namespace cache? Or are they kept separate? Or is this what you were referring to when you said there will be more auto-generated prefixes (generated by prez curie maker).

Yes as part of the prefix retrieval if there is not yet a prefix one will be generated. https://github.com/RDFLib/prez/blob/1cdf66c355148a6230bf7d40dbef357de3fc6332/prez/services/curie_functions.py#L90 There is both:
1. A prefix namespace mapping "cache" in Prez, which is an rdflib namespace manager instance, so that the compute_qname and expand_curie can be used
2. An aiocache memory cache of curie instances to their full IRIs

prez/reference_data/profiles/prez_default_profiles.ttl
```diff
@@ -109,11 +107,14 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
                              skos:Collection ,
                              prez:CQLFilterResult ,
                              prez:QueryablesList ,
+                             sdo:DataCatalog ,
```
ashleysommer, line 110: Is this addition supposed to be in this PR?
Should there be a separate PR to add `sdo:DataCatalog`, `sdo:Dataset`, and `sdo:hasPart/isPartOf` support into prez default profiles?

recalcitrantsupplant: It was intentional for testing but I didn't think it through and will roll it back - otherwise I should roll out the SDO changes generally and I would prefer not to have the default endpoints/profiles cover all the alternate paths, it makes the endpoint definitions harder to read, and the queries more complex, and in most cases I think people should pick one or the other (if possible). I would rather put the effort into providing examples or making it easier for people to choose one or the other but not both.

PR #403, opened 2025-07-15
fix: for missing ?focus_node var on inner subselect for CQL-genreated SUBSELECT query

### LINKED ISSUE #402: ?focusNode is not in list of vars in CQL-generated sub-SELECT query
Looks like #387 fixes some search query generation, but broke CQL queries. Not sure why this isn't picked up by the test suite.

I'm seeing this query generated by my deployed Prez instance
```sparql
WHERE {
  {
    SELECT DISTINCT ?path_node_2
    WHERE {
      ?focus_node <http://rs.tdwg.org/dwc/iri/toTaxon>/(<http://rs.tdwg.org/dwc/terms/acceptedNameUsageID>|<http://rs.tdwg.org/dwc/terms/parentNameUsageID>) ?path_node_2 .

      VALUES ?path_node_2 { <https://id.biodiversity.org.au/instance/apni/7646321>  }
    } LIMIT 10 OFFSET 0
  }

  {
    ?focus_node ?preds ?vals .
  }
```
It should be `SELECT DISTINCT ?focus_node`

ashleysommer:
Fixes #402 

This adds `?focus_node` as a static always-present Variable for the inner SUBSELECT on the CQL query generation.

This is required because #387 removed the addtion of "?focus_node" from the umbrella query.

As @recalcitrantsupplant mentioned in #387, its better to have the Search and CQL query generation include `?focus_node" themselves rather than add it in the umbrella, so the subquery is complete in and of itself, for testing purposes.

Evidently there was no testing for this, because the CQL query breakage flew under the radar.

### DIFF
--- a/prez/services/query_generation/cql.py
+++ b/prez/services/query_generation/cql.py
@@ -92,7 +92,8 @@ class CQLParser:
             queryable_props=None,
     ):
         self.inner_select_gpntotb_list: list[GraphPatternNotTriples] = []
-        self.inner_select_vars: list[Var] = []
+        # Always include at least ?focus_node SELECT var
+        self.inner_select_vars: list[Var] = [Var(value="focus_node")]
         self.cql: dict = cql
         self.cql_json = cql_json
         self.var_counter = 0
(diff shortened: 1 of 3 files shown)


recalcitrantsupplant (approved):
looks good thank you

PR #406, opened 2025-07-23
fix: cql and queryables updates

recalcitrantsupplant:
fix: add missing cql <> operator; add loading of local queryable definitions
fix: bugfix for specific case where a CQL filter is used with operand is "=" and the property is a SHACL Defined queryable and the object is an IRI. This didn't produce a filter, now it does. When operand is "=" but not a SHACL defined queryable, it just does a triple pattern match. When SHACL defined queryable and object is not an IRI or operand is not "=", a FILTER is always used.

### DIFF
--- a/prez/app.py
+++ b/prez/app.py
@@ -50,7 +50,7 @@ from prez.services.app_service import (
     prefix_initialisation,
     retrieve_remote_jena_fts_shapes,
     retrieve_remote_queryable_definitions,
-    retrieve_remote_template_queries,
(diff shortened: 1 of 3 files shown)


recalcitrantsupplant:
@lalewis1 are you able to re-review, NB I've refactored the SPARQL query generation to a separate helper module, and added a bunch of type hints, these changes are not intended to change any functionality.

lalewis1 (approved):
lgtm

Issue #405, opened 2025-07-21
Query optimisation bug in GraphDB spatial generation


ashleysommer:
PR #373 introduced the ability for prez to produce spatial filters that employ GraphDB-specific GeoSPARQL magic predicates that utilise the GraphDB geospatial indexes (opposed to the `geof` functions).

I've observed an optimisation issue that occurs in some cases, that causes very slow query responses.

Consider the following query that is generated by using the bbox filter in the OGC Features API:
```sparql
PREFIX schema: <https://schema.org/>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX geo: <http://www.opengis.net/ont/geosparql#>
SELECT ?focus_node
WHERE {
 ?focus_node rdf:type geo:Feature .
 ?focus_node geo:hasGeometry ?geom_bnode .
 ?focus_node schema:isPartOf <https://linked.data.gov.au/dataset/bdr/9ea6c688-c6d6-4326-a69f-e23cb72d2010> .
 ?geom_bnode geo:asWKT ?geom_var_bbox .
 ?geom_var_bbox geo:sfWithin "POLYGON ((109.237579595529640 -50.303853487643790, 109.237579595529640 -7.908225597373977, 154.234511823706730 -7.908225597373977, 154.234511823706730 -50.303853487643790, 109.237579595529640 -50.303853487643790))"^^geo:wktLiteral .
}
```
The goal of this is to find all focus_nodes that are `geo:Feature`s that are within the given Dataset, and its geometry is within the given bounding box.

In this case, the bounding box is all of Australia, containing over 34 million Geometries. The given Dataset contains approx 100 Features.

Using onto:explain we can see this produces an optimisation like this:
```sparql
{
  { # ----- Begin optimization group 1 -----   
    ?geom_var_bbox geo:sfWithin "POLYGON ((109.237579595529640 -50.303853487643790, 109.237579595529640 -7.908225597373977, 154.234511823706730 -7.908225597373977, 154.234511823706730 -50.303853487643790, 109.237579595529640 -50.303853487643790))"^^<http://www.opengis.net/ont/geosparql#wktLiteral> . # FROM PLUGIN GeoSPARQL # ESTIMATED NUMBER OF ITERATIONS: 0.1
[...]

PR #410, opened 2025-08-15
feat: wrap cql datetime and bounding box expressions in filter exists

recalcitrantsupplant:
This change is specific to focus node selection so only applies to listing endpoints.
It wraps each of any:
~1. CQL expression (posted to /cql, or as a query parameter on any listing endpoint via the filter query param)~
~2. `datetime` query parameter expression~
3. `bbox` query parameter expression - only in the case of GraphDB

This is intended to improve the performance of focus_node selection, as triplestores (may) provide a faster response where the results of the expression no longer need to be returned.

To support this change an additional check has been added such that, if the **only** contents of the focus node subselect would be a `FILTER EXISTS` expression, an additional focus node class selection triple is added, which will match the default class selection triple used in property selection (`?focus_node rdf:type ?prof_1_node_1`). This ensures that at least one value is bound in the subselect.
Edit: This change should no longer be needed for FILTER EXISTS but remains for the FILTER NOT EXISTS case

An additional feature was added to support negation in CQL, i.e. the CQL "not" operator. This was implemented by wrapping child arguments in a `FILTER NOT EXISTS` expression.

Addresses https://github.com/RDFLib/prez/issues/405

### DIFF
--- a/prez/models/query_params.py
+++ b/prez/models/query_params.py
@@ -41,7 +41,7 @@ def reformat_bbox(
         if len(coords) not in [4, 6]:
             raise HTTPException(
                 status_code=400,
-                detail=f"Invalid bbox format: Expected 4 or 6 coordinates, got {len(coords)}"
+                detail=f"Invalid bbox format: Expected 4 or 6 coordinates, got {len(coords)}",
             )
 
         return coords
@@ -50,7 +50,7 @@ def reformat_bbox(
         raise HTTPException(
             status_code=400,
             detail=f"Invalid bbox format: Unable to parse coordinates as numbers. Expected format: 'minx,miny,maxx,maxy'"
-                   f", received '{bbox[0]}'"
+            f", received '{bbox[0]}'",
         )
 
 
@@ -195,7 +195,7 @@ class ListingQueryParams:
         self.profile = profile
         self.page = page
         self.limit = limit
-        self.offset = offset if not hasattr(offset, 'default') else offset.default
+        self.offset = offset if not hasattr(offset, "default") else offset.default
         self.facet_profile = facet_profile
         self.bbox = bbox
         self.filter_lang = filter_lang
@@ -206,30 +206,36 @@ class ListingQueryParams:
         self._filter = _filter
         self.mediatype = mediatype
         self.subscription_key = subscription_key
-        self.startindex = startindex if not hasattr(startindex, 'default') else startindex.default
+        self.startindex = (
+            startindex if not hasattr(startindex, "default") else startindex.default
+        )
         self.validate_pagination_params()
         self.validate_filter()
 
     def validate_pagination_params(self):
         """Validate mutually exclusive pagination parameters."""
         from fastapi import HTTPException
-        
+
         # Debug print to see what values we have
-        print(f"DEBUG - page: {self.page}, offset: {self.offset}, startindex: {self.startindex}")
-        
+        print(
+            f"DEBUG - page: {self.page}, offset: {self.offset}, startindex: {self.startindex}"
+        )
+
         pagination_params = [
             (self.page != 1, "page"),  # page=1 is default, so only conflict if changed
             (self.offset is not None, "offset"),
-            (self.startindex is not None, "startindex")
+            (self.startindex is not None, "startindex"),
         ]
-        
-        provided_params = [(provided, name) for provided, name in pagination_params if provided]
-        
+
+        provided_params = [
+            (provided, name) for provided, name in pagination_params if provided
+        ]
+
         if len(provided_params) > 1:
             param_names = [name for _, name in provided_params]
             raise HTTPException(
-                status_code=400, 
-                detail=f"Mutually exclusive pagination parameters provided: {', '.join(param_names)}. Use only one of: page, offset, or startindex."
+                status_code=400,
+                detail=f"Mutually exclusive pagination parameters provided: {', '.join(param_names)}. Use only one of: page, offset, or startindex.",
             )
 
     def validate_filter(self):
--- a/prez/services/query_generation/spatial_filter.py
+++ b/prez/services/query_generation/spatial_filter.py
@@ -389,7 +389,7 @@ def generate_bbox_filter(
                 ),
                 TriplesSameSubjectPath.from_spo(
                     geom_bn_var, IRI(value=GEO.asWKT), geom_lit_var
-                )
+                ),
             ]
         )
 
@@ -400,7 +400,9 @@ def generate_bbox_filter(
                 triples=TriplesSameSubjectPath.from_spo(
                     geom_bn_var,
                     IRI(value=GEO.sfIntersects),
-                    RDFLiteral(value=processed_wkt, datatype=IRI(value=str(GEO.wktLiteral))),
+                    RDFLiteral(
+                        value=processed_wkt, datatype=IRI(value=str(GEO.wktLiteral))
+                    ),
                 )
             )
         )
@@ -418,7 +420,9 @@ def generate_bbox_filter(
             target_system=target_system,
         )
         if not spatial_filter_gpnts:
-            raise ValueError("generate_spatial_filter_clause returned no patterns for GeoSPARQL bbox.")
+            raise ValueError(
+                "generate_spatial_filter_clause returned no patterns for GeoSPARQL bbox."
+            )
 
         for gpnt in spatial_filter_gpnts:
             ggps.add_pattern(gpnt)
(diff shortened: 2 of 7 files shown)


prez/services/query_generation/spatial_filter.py
```diff
@@ -331,38 +331,64 @@ def generate_bbox_filter(
...
+                    IRI(value=GEO.sfIntersects),
+                    RDFLiteral(value=processed_wkt, datatype=IRI(value=str(GEO.wktLiteral))),
+                )
             )
         )
 
     elif target_system in ["geosparql", "qlever"]:
-        filter_gpnts = generate_spatial_filter_clause(
+        # Add geometry triples first
+        if target_system == "geosparql":
```
ashleysommer, line 364: Are these same geometry triples already added above, at line 339?

test_data/cql/expected_generated_queries/additional_temporal_disjoint_instant.rq
```diff
@@ -2,6 +2,9 @@ CONSTRUCT {
 ?focus_node <ex:updated_at> ?dt_1_instant
 }
 WHERE {
+
+FILTER EXISTS{
 ?focus_node <ex:updated_at> ?dt_1_instant .
 FILTER (?dt_1_instant > "2012-08-10T05:30:00+00:00"^^<http://www.w3.org/2001/XMLSchema#dateTime> || ?dt_1_instant < "2012-08-10T05:30:00+00:00"^^<http://www.w3.org/2001/XMLSchema#dateTime>)
+}
```
ashleysommer, line 9: Should these examples have the fallback triple added?
`?focus_node rdf:type ?prof_1_node_1`  
Or is this testing only the CQL generated filter, and the fallback triple gets added in `umbrella.py`?

ashleysommer:
@recalcitrantsupplant have you done any performance testing on these changes?
I think there could be cases where _always_ wrapping _all_ CQL filters in `FILTER EXISTS` could make things slower due to changing the query optimisation groups.

Consider the following:

```sparql
SELECT ?focus_node WHERE {
    ?focus_node rdf:type ex:Feature .
    ?focus_node ex:startedAt ?startTime .
    FILTER (?startTime > "2012-08-10T05:30:00+00:00"^^<http://www.w3.org/2001/XMLSchema#dateTime>)
}
```
If there are a Million instances of `ex:Feature` in the graph, but only 200 of them have a `ex:startedAt` predicate, the SPARQL execution engine will optimise it to use the PSO index so it will first get all triples with predicate `ex:startedAt`, and then intersect that with all triples with class `ex:Feature`, then finally check those results against the FILTER.

VS 
```sparql
SELECT ?focus_node WHERE {
    ?focus_node rdf:type ex:Feature .
    FILTER EXISTS {
      ?focus_node ex:startedAt ?startTime .
      FILTER (?startTime > "2012-08-10T05:30:00+00:00"^^<http://www.w3.org/2001/XMLSchema#dateTime>)
    }
}
```
In this new one, it will get all one million instances of `ex:Feature` and for each one will exectute the FILTER EXISTS check on it.
As stated in the OP, this second one has the benefit that it doesn't add the `?startTime` to the result row in the intermediate table. So perhaps it evens itself out.

prez/services/query_generation/spatial_filter.py
recalcitrantsupplant: Yes - removed now

recalcitrantsupplant:
> have you done any performance testing on these changes

No I haven't - based on your above results I think it will need testing across a few common triplestores.
Is there an expected heuristic we think might be performant generally?
- does graphwise support the geosparql filter functions yet? (I think not?)
- could we move all BGPs outside the FILTER EXISTS (wouldn't help with the geo queries)
- could we move all triples where `?focus_node` is not the subject outside of the FILTER EXISTS

In the original issue specific to spatial data it was dependent on the data size unfortunately, so would the appropriate target here be what is _generally_ the fastest?

prez/services/query_generation/spatial_filter.py
recalcitrantsupplant: in https://github.com/RDFLib/prez/pull/410/commits/e1cb3c77d65a48af6f63c8d1a8ced69417768e0d

prez/models/query_params.py
```diff
@@ -188,9 +206,33 @@ def __init__(
         self._filter = _filter
         self.mediatype = mediatype
         self.subscription_key = subscription_key
-        self.startindex = startindex
+        self.startindex = startindex if not hasattr(startindex, 'default') else startindex.default
+        # Temporarily disable pagination validation for testing
+        # self.validate_pagination_params()
```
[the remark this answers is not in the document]
recalcitrantsupplant: in https://github.com/RDFLib/prez/pull/410/commits/e1cb3c77d65a48af6f63c8d1a8ced69417768e0d

ashleysommer:
> 1. does graphwise support the geosparql filter functions yet? (I think not?)

They do work, but they do not use the GeoSPARQL indexes.  Only the predicate-style comparison uses the GeoSPARQL indexes for accelerated spatial operations.

> 2. could we move all BGPs outside the FILTER EXISTS (wouldn't help with the geo queries)

When we first discussed this issue, I thought that was our original idea to solve it, and the intention of this fix.
- Keep all basic triple patterns outside of the FILTER EXISTS
- Move any time-consuming comparisions into a FILTER EXISTS (so the query executor can break on first truth, and it doesn't need to keep all the results in the result row)
- This has the bonus effect to ensure all time-consuming comparisons are run _last_ in the subquery, and those operations do not affect the query planner's optmisation.

> 3. could we move all triples where ?focus_node is not the subject outside of the FILTER EXISTS

I like this idea, but I think you've worded it backwards.
The solution would be to put any pattern where the `?focus_node` _is_ the subject outside the FILTER EXISTS, and any pattern with a different subject goes _inside_ the FILTER EXISTS.

----

Whether we go with Option 2 or Option 3, a small optimisation can be made.
- If the only item inside a FILTER EXISTS is a FILTER function, then it doesn't need to be wrapped in FILTER EXISTS.

Eg, 
```sparql
?focus_node rdf:type ex:Occurrence .
?focus_node schema:temporal ?myTime .
FILTER EXISTS {
  FILTER (?myTime > "2024-01-01T05:30:00"^^<http://www.w3.org/2001/XMLSchema#dateTime>)
}
```    
can be simplified to:
```sparql
?focus_node rdf:type ex:Occurrence .
?focus_node schema:temporal ?myTime .
FILTER (?myTime > "2024-01-01T05:30:00"^^<http://www.w3.org/2001/XMLSchema#dateTime>)
```

recalcitrantsupplant:
Great, I've got changes I'm part way through locally that are for option 3 - all triples with ?focus_node as the subject go outside the filter exists (which I did word backwards above yes).

> When we first discussed this issue, I thought that was our original idea to solve it, and the intention of this fix.

It probably was - I think when I later looked at negation I thought we could just invert any FILTER EXISTS, which conceptually works _if_ the FILTER EXISTS wraps all triples / filters.

My understanding now would be that the FILTER EXISTS need only be introduced for performance, but the FILTER NOT EXISTS would simply match any negation statements in the CQL.

ashleysommer:
@recalcitrantsupplant 
Is there an ETA on the updated version of this PR?

recalcitrantsupplant:
Regarding the geospatial queries, it definitely sounds like there's been a behaviour change at the database level, at least in one place.

To confirm what you found: 
- wkt <-> wkt, with indexed geo predicate = 0.2s
- wkt <-> wkt with standard geof predicate = 4s possibly due to cache, else ~45s

I'm hoping to have a little benchmark tool some time in the next few months where you can run a set of queries with various options (n times, spacing, different queries).

Regarding the FILTER IN change, this was generally motivated by needing to fix boolean comparisons. The minimal testing I did showed no difference in performance compared to VALUES. We have just found a case on a large dataset where it definitely does matter (VALUES = instant, FILTER = times out). So we'll likely revert the change back to using a VALUES clause in most instances:

I've asked @lalewis1 to look into VALUES vs FILTER and any differences when doing boolean/date etc. type comparisons. I think there's some subtle differences and times you should use one over the other. My current thinking is, where the queries are equivalent, we should prefer VALUES.

ashleysommer:
> To confirm what you found:
>
> wkt <-> wkt, with indexed geo predicate = 0.2s
> wkt <-> wkt with standard geof predicate = 4s possibly due to cache, else ~45s

Note quite right.
This is the summary.

wkt <-> wkt, with indexed geo predicate = 0.2s
Geom <-> wkt, with indexed geo predicate = 45s for the first 10 tests, then down to 4.1s after that?
wkt <-> wkt with standard geof predicate = 2.1s

PR #413, opened 2025-09-01
fix: remove extra serialize line that breaks oxigraph serialize

ashleysommer:
Looks like a copy+paste error made its way into the Oxigraph changes.
When using the OGC-Featues API listing endpoint, if requesting an annotation mediatype like `anot+turtle` then Prez crashes because it tries to serialize the oxigraph ItemStore twice.


recalcitrantsupplant (approved):
looks good thank you

PR #414, opened 2025-09-01
fix: update deps and add docker build to pr tests

recalcitrantsupplant:
- adds docker build to PR tests to prevent merging PRs to main and finding the Dockerfile does not build
- updates deps which were the cause of the Dockerfile not building


lalewis1 (approved):
good idea. lgtm

PR #417, opened 2025-09-10
feat: add /functions endpoint for CQL and add a number of common func…

recalcitrantsupplant:
This PR adds support for specific focus node filtering scenarios related to the tern and sosa ontologies, with a little support for more generalised complex scenarios.

This support is implemented in accordance with the [OGC Features functions section](https://portal.ogc.org/files/96288#functions) 

The functions are detailed here in code: https://github.com/RDFLib/prez/blob/0c77328ab596e9140afd032f525ed78966fa2931/prez/services/ogc_functions_data.py#L23 and can be viewed as JSON when running Prez at the new `features/functions` endpoint.

Examples of how to call the functions have been added in [an http file](https://github.com/RDFLib/prez/blob/e53e5f438b62944e56eb7d7c56125b7666453aa6/dev/cql_functions.http)

It is intended that once users have had a chance to use these for some time, usage patterns and any pain points should be clearer. At this point an RDF declarative form could be implemented, perhaps in SHACL or templated SPARQL queries.

### DIFF
--- a/prez/services/query_generation/cql_functions.py
+++ b/prez/services/query_generation/cql_functions.py
@@ -1,0 +1,203 @@
+from sparql_grammar_pydantic import (
+    IRI,
+    GroupGraphPatternSub,
+    TriplesBlock,
+    TriplesSameSubjectPath,
+    Var,
+    VarOrTerm, GraphTerm,
+)
+
+from prez.services.query_generation.grammar_helpers import (
+    convert_value_to_rdf_term,
+    create_values_constraint,
+    create_tssp_alt_or_alt_inverse, create_tssp_sequence, create_union_gpnt_from_tssps,
+)
+
+REGISTERED_CQL_FUNCTIONS = [
+    "FOIObservationFilterDirect",
+    "FOIObservationFilterSequence",
+    "hasObservation",
+    "hasAttribute",
+    "hasAdditional"
+]
+
+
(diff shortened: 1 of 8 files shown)


prez/services/query_generation/cql_functions.py
```diff
@@ -0,0 +1,203 @@
...
+                { ?obs sosa:hasSimpleResult $arg2 }
+                UNION
+                { ?obs sosa:hasResult $arg2 }
+                UNION
+                { ?obs sosa:hasResult/rdf:value $arg2}
+                VALUES $arg2 { "val1" }
+                VALUES $arg1 { <https://example.org/vocab/prop1> }
+            }"""
+            obs_prop_var = Var(value="obs_prop")
+            result_var = Var(value="result")
```
ashleysommer, line 74: Should these vars have a counter attached to the end? So that if you have multiple hasObservation conditions, or combination of hasObservation and hasAttribute, then the vars don't alias.

ashleysommer (changes requested):

prez/services/query_generation/cql_functions.py
recalcitrantsupplant: Yes thank you I planned to do this and forgot - will add it in

recalcitrantsupplant: added in https://github.com/RDFLib/prez/pull/417/commits/b608ca76d545ae827ac869b0cbd26ddac0258749 example calls that would have overlapping variables added to the http example file

PR #421, opened 2025-09-17
chore: add black formatting requirement for PRs

recalcitrantsupplant:
Can consider ruff (faster, can do same formatting as black) if there are no objections to this


lalewis1 (approved):
good idea. lgtm. i'm happy with black.

PR #422, opened 2025-09-17
feat: add faceting on objects

recalcitrantsupplant:
Adds faceting support for objects. NB this is a less common use case than faceting on a list of results. For a project a customer wishes to facet on the children of a particular resource.

Changes:
- `_create_facets_query` and `get_facet_profile_uri_from_qsa` functions added to facet.py as they're no longer listing specific.
- adjust faceting query to accept an IRI for the focus node, and no subselect (for object case)
- update listings.py and objects.py to call the updated facet query function
- minimal regression test + example data (not currently linked)

### DIFF
--- a/prez/services/listings.py
+++ b/prez/services/listings.py
@@ -205,7 +205,7 @@ async def listing_function(
         queries.append(count_query)
     facet_profile_uri = None
     if query_params.facet_profile:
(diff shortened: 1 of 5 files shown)


recalcitrantsupplant:
documentation updated

PR #424, opened 2025-09-18
fix: issue 418

recalcitrantsupplant:
Refactored SHACL defined queryable property handling in prez/services/query_generation/cql.py:328 and prez/services/query_generation/cql.py:428 so each OR branch gets its own graph pattern; (bug was only handling TSSP patterns). Branch-specific triples now live inside the UNION instead of leaking into the
  global list.
  - Added _append_property_shape_patterns and updated queryable_id_to_tssp (prez/services/query_generation/cql.py:445, prez/services/query_generation/cql.py:580) to integrate full PropertyShape structures directly into the active branch context.
  - Added regression test based on reported issue + similar issue reported elsewhere


lalewis1 (approved):
ok I think I'm happy with this.

urge to refactor shacl.py is growing though.

PR #426, opened 2025-09-19
feat: allow no query string search

recalcitrantsupplant:
This branch allows search queries without a search term (q parameter) when filtering/faceting parameters are provided.

  Key Changes:
  - Enhanced search flexibility: Modified prez/dependencies.py:270-305 to allow empty/missing search terms when filtering parameters (facet_profile, filter, bbox, datetime) are present
  - Improved error messaging: Updated error message to guide users toward available filtering options
  - Comprehensive test coverage: Added tests/test_search_empty_term.py with 8 test cases covering both legacy behavior (requiring search term) and new behavior (allowing empty term with filters)

  Before: Search endpoint required a search term (q) parameter, returning 400 error if missing
  After: Search endpoint accepts empty/missing search term when any filtering parameter is provided, enabling pure faceting/filtering workflows without text search

### DIFF
--- a/docs/faceted_search.md
+++ b/docs/faceted_search.md
@@ -78,3 +78,23 @@ WHERE {
   }GROUP BY ?facetName ?facetValue
 }
 ```
+
+## Open Faceting
+
+Open faceting is allowed, meaning you can perform faceted search without providing a search term by using only the `facet_profile` parameter:
+
+```
+/search?facet_profile=xyz
+```
+
+When no search term is provided, Prez automatically injects a `?focus_node a ?type` triple in the subselect query. This ensures that only instances of classes are faceted on, rather than attempting to facet across all nodes in the dataset.
+
+**Performance Considerations:**
+
+Open faceting works well for smaller datasets, but performance can become an issue with larger datasets since it needs to process all class instances without any filtering constraints.
+
+For larger datasets, it is recommended to:
+- Perform a search first to narrow the result set
+- Add additional filters or search terms to reduce the scope (using parameters: filter, bbox, datetime, q)
+- Use a separate full text search/faceting implementation designed for large-scale operations such as open search or lucene
+```
--- a/prez/dependencies.py
+++ b/prez/dependencies.py
@@ -270,14 +270,36 @@ async def generate_search_query(
     endpoint_uri_type: tuple[URIRef, URIRef] = Depends(get_endpoint_uri_type),
 ):
     term = request.query_params.get("q")
+
+    # Check for filtering/faceting parameters that allow empty search terms
+    def has_filtering_params():
+        params = request.query_params
+        # Check for faceting
+        if params.get("facet_profile"):
+            return True
+        # Check for CQL filter
+        if params.get("filter"):
+            return True
+        # Check for spatial filtering
+        if params.get("bbox"):
+            return True
+        # Check for temporal filtering
+        if params.get("datetime"):
+            return True
+        return False
+
     # Check if the search term 'q' is provided
     if not term:
-        # If 'q' is missing or empty, only raise error if it's the search endpoint
+        # If 'q' is missing or empty, check if we're on the search endpoint
         if endpoint_uri_type[0] == EP["extended-ogc-records/search"]:
-            raise HTTPException(
-                status_code=400,
-                detail="Search query parameter 'q' must be provided.",
-            )
+            # Allow empty search term if filtering/faceting parameters are present
+            if has_filtering_params():
+                return None  # generate query without FTS component
+            else:
+                raise HTTPException(
+                    status_code=400,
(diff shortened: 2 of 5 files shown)


lalewis1:
This doesn't work for me locally, I end up with no select clause in the query,

endpoint: `http://localhost:8000/search?facet_profile=facets&_mediatype=application/sparql-query`

resultant query:
```sparql
CONSTRUCT {
?focus_node <https://prez.dev/type> <https://prez.dev/FocusNode> .
?focus_node <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?prof_1_node_1
}
WHERE {
?focus_node <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?prof_1_node_1 .
}
```

which then triggers an error at `listings.py:203`

```python
        subselect = copy.deepcopy(main_query.inner_select)
```

when it tries to access the inner select component.

note, facet profile in use:

```turtle
@prefix dc: <http://purl.org/dc/elements/1.1/> .
@prefix dcterms: <http://purl.org/dc/terms/> .
@prefix prez: <https://prez.dev/> .
@prefix prof: <http://www.w3.org/ns/dx/prof/> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix schema: <https://schema.org/> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .


prez:SystemGraph {

<https://prez.dev/profile/facets>
  dcterms:identifier "facets"^^xsd:token ;
  dcterms:title "default facet profile" ;
  dcterms:description "Allows faceting on the objects of various properties" ;
  sh:property [ sh:path [ sh:union ( rdf:type schema:isPartOf schema:mentions dc:type ) ] ] ;
.

}

```

recalcitrantsupplant:
Thanks - I wonder what the inner select should fall back to here, probably just a class triple `?focus_node a ?type` is most appropriate.

lalewis1:
won't that impact the set of things the facet query is then generated for though?

like what if the facet query had

```turtle
...
sh:property [ sh:path [ sh:union ( schema:keywords ) ] ] ;
...
```

and the data had

```turtle
...
ex:thing a schema:Thing ;
  schema:hasPart [
    schema:name "sub thing with no class" ;
    schema:keywords "watermelon", "sugar", "high"
  ] ;
.
...
```

would the facet query then not show any options for the `schema:keywords` facet?
because we secretly add a filter for resources with a declared class.

recalcitrantsupplant:
> won't that impact the set of things the facet query is then generated for though?

Yes, by design though, faceting is already expensive, the idea is to restrict it the properties for "instances of something".

> would the facet query then not show any options for the `schema:keywords` facet? because we secretly add a filter for resources with a declared class.

The facet profile should specify a sequence path `schema:hasPart/schema:keywords` here and then you'll get faceting on the nested results.

lalewis1:
There might be another issue, i still get the same error when trying to access the inner_select of the main query

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/laws/PycharmProjects/prez/.venv/lib/python3.12/site-packages/pydantic/main.py", line 988, in __getattr__
    return super().__getattribute__(item)  # Raises AttributeError if appropriate
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/laws/PycharmProjects/prez/prez/services/query_generation/umbrella.py", line 209, in inner_select
    .content.group_graph_patterns[0]
     ^^^^^^^
  File "/home/laws/PycharmProjects/prez/.venv/lib/python3.12/site-packages/pydantic/main.py", line 991, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'TriplesBlock' object has no attribute 'content'
```

lalewis1:
yeah seems like `prez/services/query_generation/umbrella.py:209` needs to add in a check for the result type of `.graph_patterns_or_triples_blocks[0]` because TriplesBlocks have no `.content` property.

lalewis1:
I wasn't specifying anything, so i guess anot+turtle by default

lalewis1:
I was reading The lucene facetting implementation and what they do is limit the number of facets returned and order them by frequency, i wonder if we could also do this?

like by default only return the top x facet categories by member count and then have pagination to get the next x if requested.

lalewis1:
yeah fair point, I'll just test it again on a subset of the data i was using and see if any other issues come up.

lalewis1:
ok its working on the smaller dataset.

can i just suggest a doco update to go with this change, which states the performance limitations on larger datasets, and the default `?s a ?class` behavior when no search term is provided.

lalewis1 (approved):
Happy now thanks.

PR #432, opened 2025-10-07
feat: allow merging of prez config from a local dir, mainly to suppor…

recalcitrantsupplant:
This PR allows specifying a configuration directory in the structure of prez's reference_data folder.
If running directly in python, this directory alone will be used to configure prez.
If running in the supplied dockerfile, a readonly directory can be mounted to the container.
In the container entrypoint, the mounted directory will then be merged with the default prez configuration.
This means additional profiles, endpoints etc. can be added.
If custom configuration is supplied with filenames that conflict with the default configuration files, the custom configuration files overwrite defaults.
In summary, when running using the docker image:
- custom configuration can be added in addition to default config
- default configuration can be overridden.

The directory should be mounted using the `ro` flag, for example, in a docker run command:
`-v "./prez_config/reference:/app/reference_data:ro"`
If the `ro` flag is not used, the default config from prez will be written to the volume location.

It is recommended to only mount to the `app/reference_data` directory in the container as this will work with the default user permissions in the container. 

**There are no changes to the existing remote retrieval of configuration from the data repository.** This is a good option for development.

This change is motivated by wanting prez configuration to be kept separate from instance data in production environments, and removing the need to have writes to the production database for application configuration.

### DIFF
--- a/dev/cql_functions.http
+++ b/dev/cql_functions.http
@@ -93,6 +93,28 @@ GET {{baseUrl}}/cql
     &datetime=2021-01-01T00:00:00Z/2021-01-02T00:00:00Z
 Accept: application/sparql-query
 
+### qualifiedAttribution single values
+GET {{baseUrl}}/cql
+    ?filter={
+      "op": "qualifiedAttribution",
+      "args": [
+        ["https://example.org/vocab/role/contributor"],
+        ["https://example.org/vocab/org/org2"]
+      ]
+    }
+Accept: application/sparql-query
+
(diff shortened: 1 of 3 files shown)


lalewis1:
I think that the PREZ_REFERENCE_DATA_DIR could be a setting on the settings class as per the other env variable settings.

It would require moving the logic from entrypoint.sh to python code but it could be done.

I am tempted to refactor but I think I'll just approve for now. and it can be fixed up later along with consolidation of config to a single graph.

PR #433, opened 2025-10-13
perf: filter exists on subselect type

recalcitrantsupplant:
This change wraps rdf:type triples in the focus node selection subselect in a FILTER EXISTS clause to improve the performance of these queries. This change impacts:

- Listing queries for "hierarchical" endpoints e.g. `/catalogs/...` etc.
- Links generation, triggered when using "annotated" profiles (as links use the endpoint nodeshapes to determine whether a link can be generated for a given IRI based on it's class and the classes of any "parent" nodes specified in an endpoint nodeshape.)

Within property paths, only class declarations for the entire property shape (at the top level) are put into a filter exists clause. No change was made to the specification of classes in property paths applied to a specific path, these still emit rdf:type triples without a filter exists clause. This limits the scope of code changes while still covering the most common use cases (link generation and listing queries).

Example subselect before:

```sparql
SELECT DISTINCT ?focus_node
WHERE {
  ?focus_node <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?focus_classes .
  ?focus_node ^<http://purl.org/dc/terms/isPartOf>|<http://purl.org/dc/terms/hasPart>|^<https://schema.org/isPartOf>|<https://schema.org/hasPart> ?path_node_1 .
  ?path_node_1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?path_node_classes_1 .
  VALUES ?focus_classes{ <http://www.w3.org/ns/dcat#Catalog> <https://schema.org/DataCatalog>  }
  VALUES ?path_node_classes_1{ <https://schema.org/DataCatalog> <http://www.w3.org/ns/dcat#Catalog> <http://www.w3.org/2004/02/skos/core#Collection> <https://schema.org/Dataset> <https://schema.org/Resource> <http://www.w3.org/2004/02/skos/core#ConceptScheme>  }
}
```

Example subselect after:

```sparql
SELECT DISTINCT ?focus_node
WHERE {
  ?focus_node ^<http://purl.org/dc/terms/isPartOf>|<http://purl.org/dc/terms/hasPart>|^<https://schema.org/isPartOf>|<https://schema.org/hasPart> ?path_node_1 .
  FILTER EXISTS{
    ?path_node_1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?path_node_classes_1 .
    ?focus_node <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?focus_classes .
    VALUES ?focus_classes{ <https://schema.org/DataCatalog> <http://www.w3.org/ns/dcat#Catalog>  }
    VALUES ?path_node_classes_1{ <https://schema.org/DataCatalog> <http://www.w3.org/ns/dcat#Catalog> <http://www.w3.org/2004/02/skos/core#Collection> <https://schema.org/Dataset> <https://schema.org/Resource> <http://www.w3.org/2004/02/skos/core#ConceptScheme>  }
  }
}
```

### DIFF
--- a/prez/models/query_params.py
+++ b/prez/models/query_params.py
@@ -32,9 +32,26 @@ def reformat_bbox(
         example=["113.338953078, -43.6345972634, 153.569469029, -10.6681857235"],
     )
 ) -> List[float]:
-    if bbox:
-        return [float(x) for x in bbox[0].split(",")]
-    return None
+    if not bbox:
+        return None
+
+    try:
+        coords = [float(x.strip()) for x in bbox[0].split(",")]
+
+        if len(coords) not in [4, 6]:
+            raise HTTPException(
+                status_code=400,
+                detail=f"Invalid bbox format: Expected 4 or 6 coordinates, got {len(coords)}",
+            )
+
+        return coords
+
+    except ValueError:
+        raise HTTPException(
+            status_code=400,
+            detail=f"Invalid bbox format: Unable to parse coordinates as numbers. Expected format: 'minx,miny,maxx,maxy'"
+            f", received '{bbox[0]}'",
+        )
 
 
 def parse_datetime(
@@ -178,7 +195,7 @@ class ListingQueryParams:
         self.profile = profile
         self.page = page
         self.limit = limit
-        self.offset = offset
+        self.offset = offset if not hasattr(offset, "default") else offset.default
         self.facet_profile = facet_profile
         self.bbox = bbox
         self.filter_lang = filter_lang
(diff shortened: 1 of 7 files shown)


ashleysommer:
@recalcitrantsupplant I think this might still be a problem at H2 and H3, but its definitely more pronounced in the H1 listings, as they are not bound to a known parent.

I think one potential solution would be to try to detect any Class filters on the focus_node, and move those out of the FILTER.
Eg, for triples where the subject is var (?focus_node) and the predicate is `rdf:type` and the object is ?var_class, then ensure that is not in the FILTER_EXISTS, so it can be used to help target the focus_nodes. I think most popular triplestores have an class-to-object index (in GraphDB its the EntityList) than can quickly select nodes based on their class URI. 

I'm curious, do you have any cases where the FILTER EXISTS change in this PR did speed up a query?
I know we discussed the need for this a couple of months ago, as a way of speeding up some of the BDR queries, but I can't remember which they were. And I think it was mainly for the Count queries? I'll try to find it in my notes.

PR #438, opened 2025-10-15
fix: geojson response count behaviour changes

ashleysommer:
* Only include Count query for GeoJSON responses if requestType = hits
* This includes endpoints on the OGC Features API, and the regular prez listings (if its is a GeoJSON-compatible profile) and on the /cql listing endpoint.
* Infer the numberMatched count without a count query, if this is the first page, and the number of features returned is less than the per-page limit.
* Remove any mentions of `item_graph` in `ogc_features_listing()` response, because it uses only item_store
* Removed an extra nested import HTTPException that was imported on every request in the validate_pagination_params function * Changed the print output in `validate_pagination_params()` to a proper logger debug output, so it can be suppressed.
* Bump rdf2geojson converted to v0.9.0
* Bump pyoxigraph to latest v0.5.1, and corresponding oxrdflib to v0.5.0 Re-black listings.py and renderer.py

### DIFF
--- a/.env-full-template
+++ b/.env-full-template
@@ -23,6 +23,7 @@ CURIE_SEPARATOR=":"
 ORDER_LISTS_BY_LABEL="true"
 LISTING_COUNT_LIMIT="100"
 SEARCH_COUNT_LIMIT="10"
+LISTING_COUNT_ON_DEMAND="false"  # When true, counts are only returned when resultType=hits is passed
 DISABLE_PREFIX_GENERATION="false"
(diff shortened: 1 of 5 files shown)


recalcitrantsupplant:
Thank you Ashley. I am adding a small extension to have an opt in setting which gives the same behaviour for listing endpoints, i.e. exclude counts unless requested.

recalcitrantsupplant (approved):

hjohns:
All looks good to me. Gives me flexibility to render items quickly, then request a count separately.

PR #439, opened 2025-10-15
fix: filter regex should wrap the variable in STR function

recalcitrantsupplant:
This allows filtering of strings with datatypes


lalewis1 (approved):
LGTM

PR #442, opened 2025-10-31
fix: link generation

recalcitrantsupplant:
This change fixes a regression (from https://github.com/RDFLib/prez/pull/433) with link generation where incorrect additional links are generated. Separately, multiple members links are now returned, where previously only a single link was allowed.

in link_generation.py:
- ensure all required triple/gpnt patterns are included in link generation queries. The changes in https://github.com/RDFLib/prez/pull/433 meant that the sparql in tssp_exists_list and gpnt_exists_list were not included in the link generation queries, so the matching was incorrect (too broad).
- use of the "exists" patterns has not been put in to a FILTER EXISTS clause for now for link generation queries (the patterns are included as regular pattern matches as per the first dot point). As there have been a lot of performance inconsistencies generally better testing with large datasets across triplestores is needed before being such a change is made.
- remove the check for multiple members links; all members links will now be added. @hjohns @jamiefeiss could you please confirm the UI will be able to determine the most relevant members link. Probably you can identify the members link that is a single path segment longer than the current one. e.g.
A request to: `/catalogues/cats:vocab-cat1`
Could return:
`/catalogues/cats:vocab-cat1/collections` and,
`/catalogues/cats:vocab-cat1/collections/vocabs:vocab1/items`
in which case you would use the first one.

- update documentation to clarify endpoint nodeshape design.

@ashleysommer are the links/members links in the test case I've added what you are expecting? - NB I made a small changes to the data you provided

### DIFF
--- a/tests/test_issue_236_link_generation.py
+++ b/tests/test_issue_236_link_generation.py
@@ -9,6 +9,7 @@ The fix changes link_generation.py to generate links for hierarchy level > 0
 instead of > 1, and ensures that type triples are properly filtered to avoid
 redundant constraints on focus nodes.
 """
+
 from pathlib import Path
 from unittest.mock import patch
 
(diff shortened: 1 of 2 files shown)


lalewis1 (approved):
LGTM

PR #451, opened 2025-12-22
add limit to text query

lalewis1:
Improves performance of Fuseki FTS queries by adding a limit clause directly to the FTS search pattern.

i.e.

```sparql
?focus_node text:query (<pred> "term" 11) .
```

Where 11 is calculated as (limit + offset).

Might be some side effects I haven't thought about.

### DIFF
--- a/prez/config.py
+++ b/prez/config.py
@@ -113,6 +113,9 @@ class Settings(BaseSettings):
     search_uses_listing_count_limit: bool = False
     # If True, allow a single rdfs:subClassOf hop when selecting profiles. If False, exact class only.
     profile_constraint_allow_subclass: bool = False
+    # Optional inner limit for Fuseki FTS text:query. When None, no limit is added to the text:query.
+    # When set to an integer, adds that value as a limit argument to the FTS query.
+    fts_limit: Optional[int] = None
 
     @field_validator("prez_version")
     @classmethod
--- a/prez/dependencies.py
+++ b/prez/dependencies.py
@@ -371,6 +371,7 @@ async def generate_search_query(
                 tss_list=tss_list,
                 limit=limit,
                 offset=offset,
+                fts_limit=settings.fts_limit,
             )
         else:
             raise NotImplementedError(
--- a/prez/services/listings.py
+++ b/prez/services/listings.py
@@ -7,49 +7,49 @@ import logging
 
 from fastapi.responses import PlainTextResponse
 from oxrdflib._converter import to_ox
+from pyoxigraph import BlankNode as OxiBlankNode
+from pyoxigraph import DefaultGraph as OxiDefaultGraph
+from pyoxigraph import Literal as OxiLiteral
+from pyoxigraph import NamedNode as OxiNamedNode
+from pyoxigraph import Quad as OxiQuad
 from pyoxigraph import (
     RdfFormat,
-    Store as OxiStore,
-    Quad as OxiQuad,
-    NamedNode as OxiNamedNode,
-    BlankNode as OxiBlankNode,
-    Literal as OxiLiteral,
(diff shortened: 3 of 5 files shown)


recalcitrantsupplant:
I have played with this in the past, I think we (royal we, sorry didn't record my experiments..) need some test cases to confirm behaviour. From memory:

- ~the limit here surprisingly didn't seem to improve performance much (I assume this is why you want it) - happy to be shown wrong on this one~ read your initial description
- produced "wrong" results - I think because you're limiting the lucene results _then_ joining in the graph if you set the lucene limit too low you can end up with not enough results to satisfy the outer limit.

lalewis1:
I'll run some comparisons and see if I can get different results.

recalcitrantsupplant:
> @recalcitrantsupplant is the 'wrong results' you mention to do with the counts? I ran a few searches with/without the changes and the result sets seem to be the same (although not always coming out in the same order).

Not counts being wrong as such, my concern is I don't think the implementation keeps retrieving more FTS results to satisfy outer limits i.e. I believe the FTS implementation is unaware of the other filters you apply in the query -- Lucene doesn't know about them. As the FTS search is run first (typically most performant), the top N results may then all (worst case) or partially be filtered out in the subsequent filters (basic graph patterns & spatial). One example scenario this could happen with is type (rdf type) filters. If I have borehole identifiers that are all like "bore123" but I want to search for "bore" in non borehole objects, lets say names of reports, the FTS query for "bore" might return limit=1000 borehole IDs, and then your BGP would filter these all out as they're the wrong class. Generally this will happen with CQL, bbox, and datetime filters, as these all filter the focus node selection.

We should be able to set up some test cases to confirm the behaviour.

Something like (untested):

```turtle
@prefix ex:   <http://ex/> .
@prefix rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .

ex:id1 rdf:type ex:Identifier ; rdfs:label "hole123" .

ex:desc1 rdf:type ex:Description ; rdfs:label "the copper was found in hole123" .
```

Searching for hole123 should score the identifier higher.
If you then add a lucene limit=1 and type filter = ex:Description and get zero results, this would be the problem.

lalewis1:
hmm, yes ok makes sense that this would happen.

I wonder then what are the options:

1. Never limit the FTS query
    - \+ simple
    - \- reduced performance

2. Submit the query multiple times until either the inner FTS limit is unfilled (no more results) or outer limit is filled (got enough results)
    - \- complexity goes way up
    - \+ maybe improves performance (?)

3. ?

My gut tells me that 2. is a bad idea, and I'm not even sure how it would conflict the count query.

But if we stick with 1. then I need to come up with another way to boost the performance so that these search queries over large datasets don't time out.
  

edit: just realised that there is no offset parameter for the FTS limit so 2. would likely not even improve performance as you would have to keep requesting results from the start each time.

recalcitrantsupplant:
@lalewis1 I think we can just do the changes I've made - no listings.py changes required - could you please review

lalewis1:
Does this not break the count query though? Or are we not worried about that.

also, it doesn't handle pagination if the user has clicked through to the end of the result set and has requested the next page?

recalcitrantsupplant:
Discussed with Lawson, have added offset into the FTS search limit, have not explicitly modified the count query, the thinking being if the count query uses a higher FTS limit this will be the bottleneck, so may as well use the same FTS limit in the main query and the FTS limit anyway. Count queries will be not necessarily be accurate when using the FTS limit if the results returned from FTS are subsequently filtered/there are sparse joins to other graph patterns etc.

PR #458, opened 2026-03-23
feat: integrate lucene-backed /cql into listings and enrich queryables

edmondchuc:
## Summary

Adds a feature-flagged Lucene-backed `/cql` implementation integrated into Prez’s normal listing/query/render pipeline, preserves Lucene search semantics in rendered responses, enriches Lucene-derived `/queryables` metadata, and adds startup loading of queryables from a configured Jena assembler.

Also keeps a management endpoint that transforms a Jena assembler into synthetic `cql:Queryable` RDF for inspection/debugging.

Full implementation details are documented in `docs/2026-03-23-lucene-cql-queryables.md`.

## What Changed

- split legacy and Lucene-backed `/cql` routers so only one is registered at a time
- added Lucene-specific `/cql` request parsing and validation for:
  - optional `q`
  - optional `filter`
  - optional GET/POST `facets`
  - IRI-only `filter.args[].property`
- replaced the old raw Lucene SPARQL JSON path with Lucene query generation wired into the normal listing pipeline
- fixed Lucene SPARQL generation and remote execution:
  - configured Lucene index name
  - correct property-function argument order
  - valid rendered SPARQL syntax
  - better remote SPARQL error handling
- `application/sparql-query` now returns all generated queries with `# Query N` headers
- tightened test settings isolation so shared fixtures and direct `Settings(...)` constructions do not inherit unrelated Lucene config
- added `testcontainers` as a dev dependency to support the Lucene/Fuseki integration test coverage against a real containerized Jena runtime

## Lucene `/cql`

Under the feature flag, `/cql` now keeps normal Prez behavior:
- mediatype negotiation
- profile negotiation
- annotation retrieval
- link generation

while preserving Lucene semantics in the rendered response:
- hit score via `prez:searchResultWeight`
- total hits via `prez:count`
- facet rows via `prez:facetName`, `prez:facetValue`, and `prez:facetCount`

## Queryables

Lucene-derived queryables exposed through the existing OGC Features `/queryables` endpoints now include richer metadata:

- `x-prez-facetable`
- `x-prez-sortable`
- `x-prez-default-search`
- `x-prez-multi-valued`
- `x-prez-stored`
- `x-prez-indexed`
- `x-prez-lucene-field-type`

The Jena assembler transform emits these Lucene field flags into synthetic `cql:Queryable` RDF and startup loading can now generate queryables directly from a configured assembler via `jena_assembler_path`.

Startup queryables loading now:
- parses the configured assembler file at startup
- uses `jena_fuseki_dataset_name` to select the `fuseki:Service`
- merges generated, remote, and local queryables
- resolves conflicts by `dcterms:identifier` with precedence `local > remote > generated`

## Management Endpoint

Added:

- `POST /jena-assembler-to-queryables`

It accepts Turtle Jena assembler input, returns generated synthetic queryables as Turtle, and does not mutate the running instance.

This endpoint is now explicitly a pure transform for inspection/debugging; normal runtime loading uses `jena_assembler_path` at startup.

## Testing

Validated with:

- `uv run pytest tests/test_cql_lucene.py tests/test_cql.py tests/test_remote_sparql_timeout.py tests/test_management_jena_queryables_transform.py tests/test_ogc_features_manual.py`
- `uv run pytest tests/test_management_jena_queryables_transform.py tests/test_cql_lucene.py tests/test_queryables_startup.py`
- `poetry run pytest -q`

## Notes

- the Lucene-backed `/cql` path remains feature-flagged
- startup queryables generation from `jena_assembler_path` is independent from the Lucene `/cql` feature flag
- `LatLonField` is still skipped by the current queryables transform


docs/2026-03-23-lucene-cql-queryables.md
```diff
@@ -0,0 +1,251 @@
...
+
+When the feature flag is disabled, `/cql` keeps its existing behavior.
+
+When the feature flag is enabled, `/cql`:
+
+- accepts raw Lucene-backed `q`, `filter`, and `facets` inputs
+- validates that CQL `property` values are IRIs
+- executes raw SPARQL with `luc:query` and optional `luc:facet`
+- always returns `application/sparql-results+json`
+- ignores Prez content negotiation and profile negotiation for this endpoint
```
recalcitrantsupplant: we'll still need the mediatype/profile negotiation, the changes will generally be needed at all endpoints (/search, listing endpoints)

docs/2026-03-23-lucene-cql-queryables.md
```diff
@@ -0,0 +1,251 @@
...
+- or `$PREZ_REFERENCE_DATA_DIR/queryables/` when `PREZ_REFERENCE_DATA_DIR` is set
+
+Prez loads `*.ttl` and `*.rdf` files from that directory at startup.
+
+Each synthetic queryable should use the Lucene field IRI as both:
+
+- the RDF subject
+- the `dcterms:identifier` string literal value
+
+Example:
```
recalcitrantsupplant, line 179: I think we add the jena config.ttl as another source of queryable definitions, and we'll need to expose/translate probably all or most things that are now possible in the jena lucene config, as FE clients will need to know e.g. is this keyword or FTS indexed, is it range facetable etc. Could you check with @hjohns is the demo UI parsing the updated config.ttl to know how to render different filters etc.?

The CQL spec only specifies the JSON format, so we do need to support it, but prez clients aren't using it so it's lower priority, + it may be difficult to map

docs/2026-03-23-lucene-cql-queryables.md
```diff
@@ -0,0 +1,251 @@
...
+
+```json
+{
+  "x-prez-facetable": true
+}
+```
+
+For v1, global and local queryables return the same loaded dataset-derived set.
+
+## Jena Assembler Transform Endpoint
```
recalcitrantsupplant, line 234: why not just make this a startup task to parse from the config.ttl, should be pretty cheap translation/small in memory triples for the queryables

docs/2026-03-23-lucene-cql-queryables.md
recalcitrantsupplant: OK, read the bit below about the transform endpoint below, so that comment is probably the more relevant one now

prez/models/ogc_features.py
```diff
@@ -227,6 +227,7 @@ class QueryableProperty(BaseModel):
     items: Optional[Union[List[str], List[int]]] = Field(
         None, description="Items for array properties"
     )
+    x_prez_facetable: Optional[bool] = Field(None, alias="x-prez-facetable")
```
recalcitrantsupplant, line 230: I think we'll need to extend this, range facetable, whether it's a keyword field vs full text.. basically all (or most) of the info on how something was indexed might be required by a FE client to know. Referring back to https://github.com/RDFLib/prez/pull/458/changes#r2973205160 maybe check how Howard is currently rendering the UI, does it use the config/should it

prez/routers/cql_lucene_router.py
recalcitrantsupplant, line 1: Not sure how this will work without redoing a lot of other prez components. The main missing pieces are:

- profile expansion
- annotation retrieval
- link generation

Would need to be a CONSTRUCT rather than SELECT too for prez lib to parse it.

I think it will be easier to create a new query based on the existing SearchQueryFusekiFTS query in prez and modify the inputs/outputs of this to align to the new luc:query / luc:facet, then wire the relevant bits of this into umbrella.py, should be the same/similar productions that SearchQueryFusekiFTS has (lists of TSSP, GPNT)

recalcitrantsupplant (changes requested):
see inline comments

docs/2026-03-23-lucene-cql-queryables.md
edmondchuc: Addressed in https://github.com/RDFLib/prez/pull/458/commits/8aad5535ec1a151f4301fe282794e824bc000e7f

prez/models/ogc_features.py
edmondchuc: Range facetable doesn't exist in the implementation, but this commit https://github.com/RDFLib/prez/pull/458/commits/8aad5535ec1a151f4301fe282794e824bc000e7f has added other fields that are available.

prez/routers/cql_lucene_router.py
edmondchuc: Addressed in https://github.com/RDFLib/prez/pull/458/commits/8aad5535ec1a151f4301fe282794e824bc000e7f

docs/2026-03-23-lucene-cql-queryables.md
edmondchuc: Addressed in https://github.com/RDFLib/prez/pull/458/commits/3dd8ec4e68699e233cbd0d6f648aeb470772f618

docs/2026-03-23-lucene-cql-queryables.md
edmondchuc: Resolving. See the other thread.

recalcitrantsupplant:
Could we do one combined query with the facet and query property function calls in Union branches, the current thinking is this will allow us to do a single lucene call which returns both hits and facets to jena. There is an example of this style in the jena demo folder

edmondchuc:
> Could we do one combined query with the facet and query property function calls in Union branches, the current thinking is this will allow us to do a single lucene call which returns both hits and facets to jena. There is an example of this style in the jena demo folder

Addressed in https://github.com/RDFLib/prez/pull/458/commits/e113876643e9a022299d93728e45e4a04c12ed7f.

I also added testcontainers as a dev dependency to add integration tests on a real Fuseki in https://github.com/RDFLib/prez/pull/458/commits/39e253564a48789be86636a5ff31014e52620d72.

recalcitrantsupplant:
Thanks Edmond, I think this is good for testing with the GSWA data now. Suggest we leave on branch/as open PR as current deployments are based on the gswa branch.

lalewis1:
`/cql` seems to work ok for me.

Testing with the `/search` endpoint, I can get the property filtering to work with - for instance -

```http
POST http://localhost:8000/search
Content-Type: application/json
Accept: text/turtle

{
  "filter": {
    "op": "=",
    "args": [
      { "property": "urn:gswa:config#entityTypeField" },
      "https://linked.data.gov.au/def/bore/Borehole"
    ]
  },
  "limit": 5,
  "offset": 10
}
```

but facets do not work. (not sure if they should?)

```http
POST http://localhost:8000/search
Content-Type: application/json
Accept: text/turtle

{
  "filter": {
    "op": "=",
    "args": [
      { "property": "urn:gswa:config#entityTypeField" },
      "https://linked.data.gov.au/def/bore/Borehole"
    ]
  },
  "facets": [
    "urn:gswa:config#entityTypeField",
    "urn:gswa:config#agentNameField",
    "urn:gswa:config#statusField"
  ],
  "limit": 5,
  "offset": 10
}
```

_(no error but no facets returned)_

and search queries do not work:

```http
### search-query

POST http://localhost:8000/search
Content-Type: application/json
Accept: text/turtle

{
  "q": "cane",
  "limit": 5
}
```

_no error but no results_

> All the above queries work as expected and get results on the `/cql` endpoint

lalewis1:
__Low Priority__

`text:indexes` property in the jena assembler is not supported.

recalcitrantsupplant:
The priority argument to support for faceting is facet_profile (we can additionally support facet as above and supply a list of fields), but for the change to be minimal for GSWA, and to generally minimise the number of fields that would need to be specified in an API call facet_profile should be the priority. The facet profile itself would list the lucene fields, and then prez would inject these in the SPARQL query.

the current syntax (which should remain to support non jena-lucene use):
```turtle
<https://prez.dev/profile/facet-by-type>
    dcterms:identifier "facet-type"^^xsd:token ;
    dcterms:title "Facet things by type" ;
    dcterms:description "Allows faceting by rdf:type" ;
    sh:property [ sh:path [ sh:union ( rdf:type ) ] ] .
```
should have an alternate form something like:
```turtle
<https://prez.dev/profile/facet-by-type>
    dcterms:identifier "facet-type"^^xsd:token ;
    dcterms:title "Facet things by type" ;
    dcterms:description "Allows faceting by rdf:type" ;
    lucene:flatFacets urn:field1, urn:field2 ;
    lucene:rangeFacets [
       lucene:field urn:field3 ;
       lucene:bucketBoundaries ( "null" 20 30 40 "null" )
]
```
not sure if there's a better way to express null
potentially it's fine to just make bucketBoundaries a JSON string `'[null, 2020, 2022, 2024, 2026, null]'` - I can't see where this would be problematic (not sure why/where we would need to parse the individual elements as RDF collection items). The JSON string could then be injected directly

See: https://github.com/aiworkerjohns/jena/blob/main/docs/02-sparql-api.md#range-facets for more info

lalewis1:
Yeah I'm ok with a JSON string. maybe explictly declared as `"[null, 10, 20, null]"^^rdf:JSON`. how annoying that there is no null keyword.

edmondchuc:
Am I understanding this correctly that when the config `enable_cql_jena_lucene_json` is set to true, prez uses the new jena lucene `luc:query` and `luc:facet` for both `/cql` and `/search`?

For `/search`, we'll need to do a mapping from the `facet_profile` value's profile to the CQL SPARQL equivalent. Is there such a mapping already? I haven't seen the `lucene:` predicates before.

recalcitrantsupplant:
> Am I understanding this correctly that when the config `enable_cql_jena_lucene_json` is set to true, prez uses the new jena lucene `luc:query` and `luc:facet` for both `/cql` and `/search`?

it's a bit more general, when enabled, all listing endpoints (including cql and search):
- push down CQL filters to lucene where possible
- push the q= search term to lucene
- push facet_profile fields to lucene for faceting
- push limits to lucene (need to be careful which limits, top docs vs facet)
- push sorting to lucene (lets do this one later, lower priority)

> we'll need to do a mapping from the facet_profile value's profile to the CQL SPARQL equivalent. Is there such a mapping already

No there isn't, but that's what I'm proposing above. The lucene prefix I just made up for an example, I'm not sure if it's sensible to reuse the same namespace as in the jena config, maybe not, perhaps https://prez.dev/jena-lucene/ ? not fussed

This:
```turtle
    lucene:flatFacets urn:field1, urn:field2 ;
    lucene:rangeFacets [
       lucene:field urn:field3 ;
       lucene:bucketBoundaries "[null, 20, 30, 40, null]"^^rdf:JSON
```
Would get translated to:
```sparql
# Category facets + year range buckets in one request
(?f ?v ?low ?high ?c) luc:facet ("default" "learning"
    '["urn:field1", "urn:field2", {"field":"urn:field3", "ranges":[null, 20, 30, 40, null]}]'
    10) .
```
(based on a docs example see https://github.com/aiworkerjohns/jena/blob/main/docs/02-sparql-api.md#range-facets)

lalewis1:
just noticed that

```http
POST http://localhost:8000/search
Content-Type: application/json
Accept: text/turtle

{
  "filter": {
    "op": "=",
    "args": [
      { "property": "urn:gswa:config#entityTypeField" },
      "https://linked.data.gov.au/def/bore/Borehole"
    ]
  },
  "limit": 5,
  "offset": 10
}
```

returns 5 results but `prez:count 0`

prez/repositories/remote_sparql.py
lalewis1, line 1: i'd like to move alot of this request logging into the existing middleware if possible. can leave for now though and open a new change after merge.

recalcitrantsupplant: please see https://github.com/RDFLib/prez/pull/478

recalcitrantsupplant:
changes included in https://github.com/RDFLib/prez/pull/478

PR #465, opened 2026-06-09
fix: add complete Darwin Core annotations

leskneebone:
Not sure about using language tags for labels and descriptions, there seems to be a bit of inconsistency across the repo - not sure if that's normalised in downstream systems (for ref, these additions arrive out of an audit for the Biodiversity Data Repository  https://github.com/dcceew-bdr/bdr-ops/issues/217).

### DIFF
--- a/prez/reference_data/annotations/dwc-annotations.ttl
+++ b/prez/reference_data/annotations/dwc-annotations.ttl
@@ -3,6 +3,15 @@ PREFIX schema: <https://schema.org/>
 
 <http://rs.tdwg.org/dwc/terms/>
     rdfs:label "Core terms defined by Darwin Core"@en ;
-    schema:description "This term list includes all currently valid terms that have been defined in the core Darwin Core namespace dwc:.  To comment on this schema, please create a new issue in https://github.com/tdwg/dwc/issues"@en ;
+    schema:description "This term list includes all currently valid terms that have been defined in the core Darwin Core namespace dwc:.  To comment on this schema, please create a new issue in https://github.com/tdwg/dwc/issues"
 .
 
+<http://rs.tdwg.org/dwc/terms/Occurrence>
+    rdfs:label "Occurrence"@en ;
+    schema:description "An existence of a dwc:Organism at a particular place at a particular time."@en
+.
+
+<http://rs.tdwg.org/dwc/terms/Identification>
+    rdfs:label "Identification"@en ;
+    schema:description "A taxonomic determination (e.g., the assignment to a dwc:Taxon)."@en
+.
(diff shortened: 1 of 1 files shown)


recalcitrantsupplant:
Thanks Les - looks like we've run the label/description extraction on http://rs.tdwg.org/dwc/terms.ttl which doesn't have all the labels (just the one), are you able to find if there's an rdf source for them all? I can see each term has it's own RDF under a version. http://rs.tdwg.org/dwc/terms/version/eventTime-2025-06-12.ttl

Ideally there's a single RDF source we can strip all the labels/descriptions from

recalcitrantsupplant:
For the failing check, the PR title needs to be a conventional commit https://www.conventionalcommits.org/en/v1.0.0/#summary

So something like "fix: add missing dwc labels"

leskneebone:
> Thanks Les - looks like we've run the label/description extraction on http://rs.tdwg.org/dwc/terms.ttl which doesn't have all the labels (just the one), are you able to find if there's an rdf source for them all? I can see each term has it's own RDF under a version. http://rs.tdwg.org/dwc/terms/version/eventTime-2025-06-12.ttl
> 
> Ideally there's a single RDF source we can strip all the labels/descriptions from

yes, I wasn't sure if the workflow was 'just get the whole ontology' or not. We're not sure if there is such a canonical RDF version of DWC at this stage. 

Have added the labels to bdr vocabs repo instead https://github.com/dcceew-bdr/bdr-vocabs/commit/7426e42b5f00943e91f8847a11e01bf65c39f584 - will move this PR back to draft pending further investigation.

leskneebone:
Hi @recalcitrantsupplant , managed to fetch all DWC labels now - proposed file is on this branch.

recalcitrantsupplant:
Is the preflabel / label and comment / definition duplication intentional? Would be good to have one or the other

leskneebone:
> Is the preflabel / label and comment / definition duplication intentional? Would be good to have one or the other

not really motivated. What's best? looks like many annotation files here use rdfs:label and schema:description only, although some don't feature the latter (maybe not available from the sources). Happy to fix however fits best.

recalcitrantsupplant:
Ah, they're duplicated at source
I would say drop one or the other. Perhaps just keep SKOS (prefLabel + definition)

Another issue I see is prez uses skos:definition and dcterms:description as "descriptive properties" but in DWC here they are both used and have different content. Could you decide which you think is better and drop the other, or, if you think we should make an enhancement in Prez to support three annotation fields (label / definition / description) then we should open another issue for this.

"    description_predicates: list = [
        SKOS.definition,
        DCTERMS.description,
        SDO.description,
    ]"

leskneebone:
> Ah, they're duplicated at source I would say drop one or the other. Perhaps just keep SKOS (prefLabel + definition)
> 
> Another issue I see is prez uses skos:definition and dcterms:description as "descriptive properties" but in DWC here they are both used and have different content. Could you decide which you think is better and drop the other, or, if you think we should make an enhancement in Prez to support three annotation fields (label / definition / description) then we should open another issue for this.
> 
> " description_predicates: list = [ SKOS.definition, DCTERMS.description, SDO.description, ]"

Fixed now, all have only:
- One `rdfs:label`
- One `skos:definition`

recalcitrantsupplant (approved):
lgtm

PR #478, opened 2026-09-10
feat: add Jena Lucene-backed CQL search and queryables discovery, with the sparql-grammar swap

recalcitrantsupplant:
## Summary

This PR introduces Lucene-backed CQL search capabilities to Prez, enabling full-text search integration with Apache Jena Fuseki's Lucene index. It also adds automatic queryables discovery from Jena Assembler configurations and includes comprehensive testing infrastructure.

### Core Search Implementation
- **New `SearchQueryJenaLucene` class** (`prez/services/query_generation/search_jena_lucene.py`): Generates SPARQL CONSTRUCT queries that leverage Jena's `luc:query` for Lucene-backed full-text search, with support for faceting and result ranking
- **CQL Lucene router** (`prez/routers/cql_lucene_router.py`): Dedicated endpoint for Lucene-backed CQL queries with POST support
- **Queryables transformation** (`prez/services/jena_assembler_queryables.py`): Automatically transforms Jena Assembler configurations into Prez queryables, supporting text indexes and spatial fields

### Query Generation Enhancements
- Extended `PropertyShape` with CQL support (`kind: "cql"`) and Lucene-specific attributes (`cql_filter_var`, `focus_node_classes`)
- Added `sh:in` constraint detection for optimized FILTER vs VALUES clause generation
- Enhanced spatial filter generation with BIND expressions for Jena compatibility

### Dependency and Configuration Updates
- New `get_runtime_settings()` dependency for request-scoped settings
- Added `FTSUnionContainer` for managing FTS query unions
- Extended `Settings` with profile constraint distance configuration
- New `DummySearchMarker` for distinguishing search from listing contexts

### Testing Infrastructure
- **Integration tests** (`tests/integration/jena-lucene-cql/`): Full Docker-based integration tests with Fuseki container, mining dataset, and comprehensive CQL query validation
- **Unit tests** (`tests/test_cql_lucene.py`): 1750+ lines of tests covering fake Lucene repository, query generation, and result processing
- **Management tests** (`tests/test_management_jena_queryables_transform.py`): Tests for Jena Assembler transformation with various field types and constraints
- Additional tests for pivot profiles, profile constraint subclass matching, and queryables startup

### Performance and Observability
- **Request timing middleware** (`prez/middleware.py`): Captures request/response timing and logs to CSV for performance analysis
- **Timing CSV service** (`prez/services/timing_csv.py`): Structured timing data collection for benchmarking
- Enhanced annotation service with timing instrumentation

### Documentation
- Multiple design documents covering Lucene-CQL integration, POST support, performance benchmarking, and migration guides
- Lucene escaping strategy documentation
- Pivot profile and facet profile migration notes

### Infrastructure
- Docker multi-stage build improvements
- Entrypoint script for reference data merging
- Taskfile for development workflow

## Notable Implementation Details

- Lucene queries use a sentinel-based approach for empty string handling to work around Lucene limitations
- CQL property validation ensures IRIs are used in Lucene-backed mode
- Facet results are mapped from Lucene's native format into Prez RDF representation
- Profile constraint matching now supports configurable distance limits for subclass matching
- Response header trimming for large Link headers to prevent HTTP violations

https://claude.ai/code/session_01X44DhBC7NrpCJeAaCiCvXr


prez/services/listings.py
```diff
@@ -691,8 +1050,17 @@ async def ogc_features_listing_function(
             # Item store could be None, if no queries were generated
             item_store = OxiStore()
         if annotations_store is not None:
-            # Add the annotations to the store
+            merge_start = time.perf_counter()
             item_store.bulk_extend(annotations_store)
+            merge_ms = (time.perf_counter() - merge_start) * 1000
+        elif annotations_graph is not None:
```
lalewis1: annotations_graph is undefined

should this be annotations_store?

same issue on line 1059

prez/services/listings.py
```diff
@@ -193,91 +366,101 @@ async def listing_function(
...
-                        offset_clause=OffsetClause(offset=0),
+                    limit_offset=LimitOffsetClauses.create(
+                        limit=settings.listing_count_limit, offset=0
                     )
                 ),
             )
 
             # Add the subselect to the main query's where clause
-            from sparql_grammar_pydantic import (
+            from sparql_grammar import (
```
lalewis1: both imports are unused.

GroupOrUnionGraphPattern is a redefinition of earlier import (also unused)

prez/services/listings.py
recalcitrantsupplant: Good catch — it's a real `NameError`, fixed in fb18333.

Not `annotations_store` though. `annotations_graph` was the rdflib-era name; when annotations moved to an oxigraph store everything was renamed except this one `elif`, so the branch has been dead since. Repointing it at `annotations_store` would make it unreachable, because the preceding `if annotations_store is not None` already covers the only case that can supply annotations — so I removed the fallback instead. That left `to_ox` unused, so its import went too.

It was reachable: `resultType=hits` generates no main query, so `item_store` is `None`, so `annotations_store` stays `None`, and any annotated RDF mediatype fell through to the undefined name. Added `test_ogc_features_items_hits_annotated` driving that request end to end — it raises `NameError` at line 1056 without the fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

prez/services/listings.py
recalcitrantsupplant: Removed the whole local import block in 16c5e7d.

One correction: `GroupOrUnionGraphPattern` was used, on the line right below — but as you say it was a redefinition of the module-level import at line 35, so the local one just shadowed it with the same object. Dropping the block leaves line 35 doing the work. `GraphPatternNotTriples` was genuinely unused.

I swept the rest of the file for the same pattern and found two more: `import time` inside `warm_queryables_cache` and `get_cached_queryables_response`, both shadowing the module-level import at line 7. Those are gone too.

---
_Generated by [Claude Code](https://claude.ai/code)_

lalewis1:
suggest running ruff check and something like pyrefly check or mypy for static code checking.

many fixable errors are present.

recalcitrantsupplant:
Agreed, and it would have caught both of your findings — CI runs `black` but no linter, so an undefined name compiles fine and only fails at runtime.

I ran ruff and cleared the pyflakes findings in `prez/` in 604c884: 15 unused imports, an empty f-string, two unused exception bindings, and a `perf_counter` pair in `get_annotation_properties_for_oxigraph` whose elapsed value was never logged. 19 findings, all dead weight, no behaviour change.

The reassuring part: **F821 (undefined name) is now clean across the repository**, so the `NameError` was the only one of its kind. B023 (loop-variable captured in a closure) is clean in `prez/` too — those are the two rules that catch actual bugs rather than lint.

I've stopped short of adopting the tooling in this PR, because the numbers suggest it needs its own decisions rather than being bolted onto a 137-file branch:

| | count | notes |
|---|---|---|
| `PT014` | 4277 | duplicate pytest parametrize cases, all in `tests/` — dwarfs everything else |
| `B008` | 185 | **176 are `Depends`** — the FastAPI idiom, a false positive unless `extend-immutable-calls` is configured |
| `UP006`/`UP045`/`UP035` | ~320 | typing modernisation, mostly autofixable but a large diff |
| `F401`/`F841` in `tests/` | 34 | same class as what I just fixed, not yet touched |

So adopting the default rule set as-is means 185 false positives on day one and a 4000-line signal buried under `PT014`. My suggestion would be a separate PR that adds ruff with a narrow starting set — pyflakes `F` plus `B` — configures `extend-immutable-calls` for `Depends`/`Query`/`Path`, and wires `ruff check` into `pr.yaml` alongside `black-check`; then widen the rules over time rather than all at once. `PT014` and the typing modernisation are worth their own passes.

That's the author's call rather than mine, so I'll leave it here unless @recalcitrantsupplant wants it in this branch. I haven't looked at pyrefly or mypy — this codebase is only partially annotated, so I'd expect that to be a bigger piece of work than ruff.

---
_Generated by [Claude Code](https://claude.ai/code)_

lalewis1:
agreed. more comprehensive linting / type checking can be left for another PR.

lalewis1:
should GSWA specific benchmarking documentation / results (some of which includes endpoint details) really be mainlined?

recalcitrantsupplant:
No, they shouldn't have been — and it was worse than endpoint details. Removed in aea2fe6.

`docs/2026-05-13-prez-search-fuseki-benchmark-notes.md` carried, in a public repo:

- a production hostname, `prez-prod.<...>.australiaeast.azurecontainerapps.io`
- **the public IP of a Fuseki SPARQL endpoint**, `http://20.11.43.102/gswads/sparql`
- an internal IP, `http://10.0.0.4:3030`
- four absolute paths from a developer's machine, under `/home/david/PycharmProjects/prez/`

The Fuseki IP is the one I'd treat as urgent rather than untidy — a directly addressable SPARQL endpoint, and the note says it was reachable from outside the VNet.

Nothing is lost by deleting it. Those were raw notes whose generalisable conclusions already got distilled into `docs/2026-05-15-annotation-cache-and-search-performance-design.md` — Fuseki execution is not the bottleneck, `bbox` was unreliable for the investigation, cold annotation enrichment is the remaining hotspot, `/sparql` streams where `/search` materialises. That doc has no deployment detail in it. Nothing linked to the removed file.

I also swept the rest of the branch rather than assuming it was one file. Two other things:

- `docs/2025-11-26-cql-filter-values-heuristic.md` had two more `/home/david/...` paths in its manual-testing section — now repo-relative (one pointed at a local scratch file that isn't in the repo, so it's described by what it contains instead).
- The other performance docs (`2026-05-15`, `2026-05-18`, `2026-04-14-jena-lucene-pagination-notes`) and everything under `benchmark/` — including `baseline.json` and `results/*.json` — are clean: no hostnames, no IPs, no personal paths. Those are generic Prez benchmarks and I've left them.

Two things worth flagging that I have **not** changed:

1. **Deleting the file does not remove it from history.** The IP and hostname are still in the branch's earlier commits, so a merge commit carries them into `main`. If that matters, this wants a squash merge — or treating the endpoint as exposed and rotating regardless, since it has been visible on a public PR.
2. `prez/reference_data/annotations/vann-vocab-20100607.ttl` line 9 has a `file:///home/david/...` URI, but it is **pre-existing on `main`**, not from this branch — an artifact of an rdflib conversion. Out of scope here; worth a separate tidy-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

lalewis1:
I think the logging and log to csv that's littered throughout the code should be cleaned up into the existing request logging middleware if possible.

in the interest of time though, I think it can be left for this PR and addressed later.

lalewis1:
### 1. BREAKING CHANGE: CQL POST Endpoint Body Format (CRITICAL)
**File:** `prez/dependencies.py` (lines 235-285)

**Issue:** The CQL POST endpoint body format has **completely changed** in a breaking way:
- **OLD:** Raw CQL expression as the entire body
- **NEW:** JSON object with a `filter` key containing the CQL2-JSON expression

```python
# OLD (no longer supported)
POST /cql
Body: {"op": "s_intersects", "args": [...]}

# NEW (required)
POST /cql
Body: {"filter": {"op": "s_intersects", "args": [...]}, "limit": 10}
```

**Impact:** All existing clients using the CQL POST endpoint will break.

**Recommendation:** This needs to be documented as a BREAKING CHANGE in release notes. Consider if a migration path is needed.


@recalcitrantsupplant  Is this impact known?

recalcitrantsupplant:
Confirmed against `main` — the report is accurate. Leaving the "is this intended and acceptable" call to @recalcitrantsupplant, but here are the verified specifics so the decision has them.

**What changed.** `main` passed the whole body through:

```python
cql_parser = CQLParser(cql_json=body, queryable_props=queryable_props)   # main
cql_json = body.get("filter")                                            # this branch
```

So a client posting `{"op": "s_intersects", "args": [...]}` now gets `400 Request body must contain a 'filter' key`. No fallback.

**There is a second breaking change in the same function**, not in the report above: the branch now requires `Content-Type: application/json` and returns **415** otherwise. `main` had no content-type check at all, so a client sending the correct body with a missing or different content-type breaks too — and that one is easy to miss, because the body shape is right.

**Why the wrapper exists**, for the record: POST bodies now carry the full listing parameter set next to the filter — `limit`, `offset`, `page`, `_profile`, `_mediatype`, `bbox`, `datetime`, `order_by`, `q`, `facets` and so on. Those can't be expressed alongside a raw CQL expression body, so the wrapper isn't gratuitous.

**A backwards-compatible path exists and is small,** if a clean break isn't wanted. CQL2-JSON always has a top-level `op`, and the new wrapper never does, so the two shapes are unambiguous:

```python
cql_json = body.get("filter")
if cql_json is None and "op" in body:
    cql_json = body          # pre-478 raw-expression body
```

That keeps old clients working, supports the new richer shape, and needs no version negotiation. The content-type check could be relaxed to "reject only if a non-JSON type is explicitly set" on the same reasoning.

So the options are roughly: accept the break and document it in release notes (the docstring already says BREAKING CHANGE, so it was a deliberate choice); or take the ~5-line compat shim above and drop the hard break in a later major version. Either way the content-type 415 should be called out in the notes alongside the body shape — it's a separate way for the same clients to break.

---
_Generated by [Claude Code](https://claude.ai/code)_

recalcitrantsupplant:
Confirmed with @recalcitrantsupplant: the impact is known and intended. This branch is groundwork for v5, and the changes are moving to another repository, so **no compatibility shim** — the break stands and gets documented instead.

Documented in 021880a as [`docs/2026-09-21-v5-breaking-changes.md`](docs/2026-09-21-v5-breaking-changes.md), and summarised at the top of the PR description. I swept the branch for other breaks rather than only writing up the one reported; four in total:

1. **`POST /cql` body wrapped in `filter`** — as you described, with the migration and the reason the wrapper exists.
2. **`POST /cql` requires `Content-Type: application/json`** — the second break in that function, which breaks independently of the body shape.
3. **`sparql-grammar-pydantic` → `sparql-grammar`** — affects downstream code importing Prez's query generation classes. Emitted SPARQL is unchanged except that `CollectionPath` now renders its items separated rather than concatenated.
4. **`pyld` `^2.0.4` → `^3.0.0`** — the only dependency bump that crosses a major version.

The note also records what is *not* a break, since three things read like one in the diff: `/cql` GET and POST changed router modules but kept their paths and endpoint names; the thirteen new settings are additive with no removals, renames or changed defaults; and `POST /sparql` does not go through the new JSON body parser, so the SPARQL protocol content types still work.

One thing that surfaced while writing it and is worth its own attention: **`sparql-grammar` is pinned as a git dependency at `rev = "main"`**, which cannot be published to PyPI. That needs to become a tag or a released version before a v5 release is cut. Flagged in the note and in the PR description.

---
_Generated by [Claude Code](https://claude.ai/code)_

lalewis1:
we should probably at least pin sparql-grammar to a tag or commit hash. So it doesn't break underneath us.

recalcitrantsupplant:
Agreed and done in 47296fe — pinned to `f191cb9944f08248eb883b50a8a7eb4704e42206`.

The library has no tags, so a commit is the tightest pin available. That commit is the one `poetry.lock` had already resolved `main` to, so nothing installed changes: the lock moves only its content hash, and the test suite is unchanged at 587 passing.

One wrinkle worth recording rather than hiding. The lock still reads:

```toml
reference = "main"
resolved_reference = "f191cb9944f08248eb883b50a8a7eb4704e42206"
```

Poetry reuses the cached resolution when the commit hasn't moved, so it left the `reference` string alone. I tried `poetry lock --regenerate`, which does fix the field — but it re-resolves the whole graph and moves **48 other packages** within their constraints, a 2366-line lock diff that changes what CI and production install. Not a trade I'd make for a metadata string on an already-large PR, so I reverted it.

Nothing is lost by leaving it: installs follow `resolved_reference`, which is the pinned commit, and `pyproject.toml` can no longer drift regardless. The field will correct itself the next time the lock is legitimately regenerated — ideally in a PR that reviews those 48 bumps deliberately, which is worth doing on its own.

The release note keeps its item about needing a tag: a commit pin stops it moving underneath us, but a git reference of any kind still can't be published to PyPI.

---
_Generated by [Claude Code](https://claude.ai/code)_

recalcitrantsupplant:
we can move to uv and release sparql grammar versions as separate tasks

lalewis1 (approved):
LGTM

